PR #10, opened 2024-02-07
Add support for end-to-end encrypted chats over WhatsApp

### LINKED ISSUE #7: Messenger encryption support
If you want to be notified when the feature is ready, you can subscribe to this issue.

tulir:
Fixes #7


nclm:
Thanks for this work!

The title here says WhatsApp but it’s also for Messenger, right?

If so, do you have any approximate idea of when it might be in use in Beeper? :)

maltee1:
My understanding is that messenger encrypted chats use the WhatsApp infrastructure.

PR #13, opened 2024-02-16
Retry connect

Fizzadar:
1st commit does the dumb thing, 2nd takes it further but feels like we could/should abstract a retry method out in either a util or even mautrix-go? WDYT @tulir?

### DIFF
--- a/user.go
+++ b/user.go
@@ -54,6 +54,8 @@ import (
 	"go.mau.fi/mautrix-meta/messagix/types"
 )
 
+const maxConnectAttempts = 5
+
 var (
 	ErrNotConnected = errors.New("not connected")
 	ErrNotLoggedIn  = errors.New("not logged in")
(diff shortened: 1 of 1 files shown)


user.go
```diff
@@ -569,16 +585,20 @@ func (user *User) unlockedConnectWithCookies(cookies cookies.Cookies) (*messagix
 		return nil, fmt.Errorf("failed to get proxy: %w", err)
 	}
 	user.log.Debug().Msg("Connecting to Meta")
-	// TODO set proxy for media client?
-	cli, err := messagix.NewClient(MessagixPlatform, cookies, log, proxyAddr)
+	var cli *messagix.Client
+	err = doWithRetry(maxConnectAttempts, func() error {
+		// TODO set proxy for media client?
```
tulir: the TODO shouldn't be moved

user.go
```diff
@@ -569,16 +585,20 @@ func (user *User) unlockedConnectWithCookies(cookies cookies.Cookies) (*messagix
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to prepare client: %w", err)
 	}
 	if user.bridge.Config.Meta.GetProxyFrom != "" {
 		cli.GetNewProxy = user.getProxy
 	}
 	cli.SetEventHandler(user.eventHandler)
-	err = cli.Connect()
+	err = doWithRetry(maxConnectAttempts, cli.Connect)
 	if err != nil {
```
tulir: This one doesn't need a retry loop, it should never fail here

Fizzadar:
Super quick @tulir 🙏! Updated

user.go
```diff
@@ -570,10 +572,20 @@ func (user *User) unlockedConnectWithCookies(cookies cookies.Cookies) (*messagix
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to prepare client: %w", err)
+	attempts := 1
+	var cli *messagix.Client
+	for {
+		cli, err = messagix.NewClient(MessagixPlatform, cookies, log, proxyAddr)
+		if err != nil {
+			if attempts > maxConnectAttempts {
+				return nil, fmt.Errorf("failed to prepare client: %w", err)
+			}
```
tulir, line 595: Oh right, this should also check that it's not a auth error (the ones listed in `Connect()`: token invalidated, challenge/consent required, LS error responses).

Or maybe check that it's a connection error specifically? That might be harder though

PR #52, opened 2024-04-17
Add image source docker label so tools can generate changelog for images

halkeye:
Tools like renovate and dependabot depend on docker container labels to generate changelogs

Eg. https://docs.renovatebot.com/modules/datasource/docker/

Ideally org.opencontainers.image.revision should be injected as well, but I don't really know where the CI scripts are

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -1,5 +1,8 @@
 FROM alpine:3.19
 
+# should probably be https://mau.dev/mautrix/meta but tooling can read scm data from gitlab.com and github.com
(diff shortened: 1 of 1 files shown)


tulir:
Probably not going to add any labels before figuring out how to deduplicate dockerfiles

halkeye:
I only just realised what this meant. I can deduce the docker files I think. Do you have a recommendation on which repo to submit a pr too?

tulir:
Probably needs to be in mautrix/ci

PR #57, opened 2024-05-07
Fix message replies in e2e threads

javiercr:
In an e2e encrypted conversation/thread, message replies sent through the bridge were not rendered in Messenger web/desktop (while they were rendered in the Messenger mobile app).

Debugging the service worker for messenger.com, we can compare an incoming reply message sent from the official Messenger web client (left) to one sent from mautrix/meta bridge (right):

Untitled 2

From this, we can infer that `quotedMessage.remoteJid` is not needed (which, btw, used to have the same value as `quotedMessage.participant`).

### DIFF
--- a/msgconv/to-whatsapp.go
+++ b/msgconv/to-whatsapp.go
@@ -101,8 +101,7 @@ func (mc *MessageConverter) ToWhatsApp(
 	var meta waMsgApplication.MessageApplication_Metadata
 	if replyTo := mc.GetMetaReply(ctx, content); replyTo != nil {
 		meta.QuotedMessage = &waMsgApplication.MessageApplication_Metadata_QuotedMessage{
-			StanzaID:  replyTo.ReplyMessageId,
-			RemoteJID: mc.GetData(ctx).JID().String(),
+			StanzaID: replyTo.ReplyMessageId,
 			// TODO: this is hacky since it hardcodes the server
 			// TODO 2: should this be included for DMs?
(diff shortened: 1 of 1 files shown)


msgconv/to-whatsapp.go
```diff
@@ -101,8 +101,7 @@ func (mc *MessageConverter) ToWhatsApp(
 	var meta waMsgApplication.MessageApplication_Metadata
 	if replyTo := mc.GetMetaReply(ctx, content); replyTo != nil {
 		meta.QuotedMessage = &waMsgApplication.MessageApplication_Metadata_QuotedMessage{
-			StanzaID:  replyTo.ReplyMessageId,
-			RemoteJID: mc.GetData(ctx).JID().String(),
+			StanzaID: replyTo.ReplyMessageId,
 			// TODO: this is hacky since it hardcodes the server
 			// TODO 2: should this be included for DMs?
```
javiercr, line 106: @tulir does this clear this TODO?

tulir: Probably not, the TODO is for the `Participant` field

tulir: Oh actually it might clear the second one if you've confirmed DMs are supposed to have that field. I think on WhatsApp the same field is meant for groups (and RemoteJID is used for DM recipient or group ID)

tulir (approved):

msgconv/to-whatsapp.go
javiercr: From the screenshot in the PR description it seems that the `Participant` field is indeed required.

PR #58, opened 2024-05-09
Add periodic refresh for connection

javiercr:
We've observed a number of intermittent errors where the socket connection is closed while trying to send a message (`failed to send LS request: not connected`).

Until we figure out why that's happening, this should help mitigate those problems by refreshing the connection at a given interval.

### DIFF
--- a/example-config.yaml
+++ b/example-config.yaml
@@ -95,8 +95,8 @@ meta:
     # HTTP endpoint to request new proxy address from, for dynamically assigned proxies.
     # The endpoint must return a JSON body with a string field called proxy_url.
     get_proxy_from:
-    # Interval to force refresh the connection (disconnect and re-connect), default is 5 minutes. Set 0 to disable refreshes.
-    refresh_interval_seconds: 300
+    # Interval to force refresh the connection (disconnect and re-connect), default is 1 day. Set 0 to disable refreshes.
+    refresh_interval_seconds: 86400
 
 # Bridge config
 bridge:
--- a/example-config.yaml
+++ b/example-config.yaml
@@ -95,10 +95,10 @@ meta:
     # HTTP endpoint to request new proxy address from, for dynamically assigned proxies.
     # The endpoint must return a JSON body with a string field called proxy_url.
     get_proxy_from:
-    # Interval to force refresh the connection (disconnect and re-connect), default is 1 day. Set 0 to disable refreshes.
-    refresh_interval_seconds: 86400
-    # Minimum interval between full reconnects in seconds
-    min_full_reconnect_interval_seconds: 1
+    # Minimum interval between full reconnects in seconds, default is 1 hour
+    min_full_reconnect_interval_seconds: 3600
+    # Interval to force refresh the connection (full reconnect), default is 1 day. Set 0 to disable force refreshes.
+    force_refresh_interval_seconds: 86400
 
 # Bridge config
 bridge:
--- a/example-config.yaml
+++ b/example-config.yaml
@@ -95,6 +95,8 @@ meta:
     # HTTP endpoint to request new proxy address from, for dynamically assigned proxies.
     # The endpoint must return a JSON body with a string field called proxy_url.
     get_proxy_from:
+    # Interval to force refresh the connection (disconnect and re-connect), default is 5 minutes. Set 0 to disable refreshes.
+    refresh_interval_seconds: 300
 
 # Bridge config
 bridge:
--- a/example-config.yaml
+++ b/example-config.yaml
@@ -97,6 +97,8 @@ meta:
     get_proxy_from:
     # Interval to force refresh the connection (disconnect and re-connect), default is 1 day. Set 0 to disable refreshes.
     refresh_interval_seconds: 86400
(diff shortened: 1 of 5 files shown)


example-config.yaml
```diff
@@ -95,6 +95,8 @@ meta:
     # HTTP endpoint to request new proxy address from, for dynamically assigned proxies.
     # The endpoint must return a JSON body with a string field called proxy_url.
     get_proxy_from:
+    # Interval to force refresh the connection (disconnect and re-connect), default is 5 minutes. Set 0 to disable refreshes.
+    refresh_interval_seconds: 300
```
javiercr: Should this config go here? Do we want a different default value other than 5 mins?

tulir: Probably fine to have a default as everyone likely has the problem. Default value should be a day or two

messagix/client.go
```diff
@@ -388,3 +394,16 @@ func (c *Client) GetTaskId() int {
 	c.activeTasks = append(c.activeTasks, id)
 	return id
 }
+
+func (c *Client) refreshConnectionPeriodically(interval time.Duration) {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	for range ticker.C {
```
tulir: The loop needs to be stopped if the client is permanently disconnected (e.g. gets logged out or something like that). Preferably it'd only run if there hasn't been a refresh recently, but that part is maybe not as important. It might be possible to do both by starting/stopping the loop in the connect/disconnect functions. Perhaps it shouldn't even be a loop, but just a timer on connect (that then gets started again on the new connection)

messagix/client.go
```diff
@@ -388,3 +394,16 @@ func (c *Client) GetTaskId() int {
...
+		c.Logger.Info().Msg("Refreshing connection")
+		c.Disconnect()
+		err := c.Connect()
```
tulir: This looks wrong, I think refreshing needs to call `LoadMessagesPage` (basically what the bridge `User` struct's connect does, not just reconnecting to MQTT)

It might be possible to resolve the issue root in other ways, but I'm not sure how exactly. There's a chance the error in question happens if the version ID we send in all requests is too old, so just fetching a new version ID could be enough.

javiercr: Just moved the logic to `User`

example-config.yaml
```diff
@@ -95,6 +95,10 @@ meta:
     # HTTP endpoint to request new proxy address from, for dynamically assigned proxies.
     # The endpoint must return a JSON body with a string field called proxy_url.
     get_proxy_from:
+    # Minimum interval between full reconnects in seconds, default is 1 hour
+    min_full_reconnect_interval_seconds: 3600
```
javiercr, line 99: I'm exposing MinFullReconnectIntervalSeconds in the config file, as it also affect the interval for force refreshes.

messagix/client.go
javiercr: Replaced the ticker with a simple timer that gets set after every new connection as suggested!

user.go
```diff
@@ -542,6 +541,18 @@ func (user *User) unlockedConnect() {
 		}
 		go user.sendMarkdownBridgeAlert(context.TODO(), "Failed to connect to %s: %v", user.bridge.ProtocolName, err)
```
tulir, line 543: There needs to be a return here to avoid starting the refresh timer if the connection failed

Also, the timer still doesn't get stopped if the client got logged out. I think stopping the timer in the `*messagix.Event_PermanentError` handler would be enough to handle that issue

javiercr: Good catch! Done!

user.go
```diff
@@ -541,6 +541,19 @@ func (user *User) unlockedConnect() {
 			})
 		}
 		go user.sendMarkdownBridgeAlert(context.TODO(), "Failed to connect to %s: %v", user.bridge.ProtocolName, err)
+		return
+	}
+
+	refreshInterval := time.Duration(user.bridge.Config.Meta.ForceRefreshIntervalSeconds) * time.Second
+	if refreshInterval > 0 {
+		user.log.Debug().Msgf("Connection will be refreshed at %s", time.Now().Add(refreshInterval).Format(time.RFC3339))
```
tulir: Msgf is not allowed (that's why the CI is angry), all log variables should be added in separate fields. Installing [pre-commit](https://pre-commit.com/) is also recommended to catch linting issues locally

user.go
```diff
@@ -1153,6 +1166,7 @@ func (user *User) eventHandler(rawEvt any) {
 		user.BridgeState.Send(user.metaState)
 		go user.sendMarkdownBridgeAlert(context.TODO(), "Error in %s connection: %v", user.bridge.ProtocolName, evt.Err)
 		user.StopBackfillLoop()
+		user.forceRefreshTimer.Stop()
```
tulir:
```suggestion
		if user.forceRefreshTimer != nil {
			user.forceRefreshTimer.Stop()
		}
```

tulir (approved):
Looks like it might work now, just a couple small tweaks

user.go
javiercr: Just noticed that, fixed it already :)

PR #63, opened 2024-05-19
Improve sending messages while reconnecting

javiercr:
# Context

Now that #58 has been merged, the connection will be refreshed (reconnected) periodically. The challenge now is to ensure that messages sent by the user during the reconnection are not lost but are properly delivered to the recipient.

To better illustrate the problem here are are some diagrams:

image
image

Colored in red are the changes we want to make to the current implementation.

# Description

This PR adds a `CanSendMessages` boolean flag to the `Client` struct, accompanied by a `SendMessagesCond` ([sync.Cond](https://pkg.go.dev/sync#Cond)).

Once the initial bootstrapping after a new connection is established is completed, we receive an `Event_Ready`. There we set the new `CanSendMessages` flag, and notify the change using `SendMessagesCond.Signal()`.

Before we send any new message to Meta, in `handleMatrixMessage`, we make sure to wait for the flag to be true using `SendMessagesCond.Wait()`. This way, `handleMatrixMessage` will be blocked while the reconnection is happening, effectively resuming once we're ready to send the messages.

Finally, note that setting `CanSendMessages` to false when disconnecting is not needed because the `Client` object is already [disposed](https://github.com/mautrix/meta/blob/6acd8d2a1c8d79c5a22b48f068282b402209e745/user.go#L1194)  (set to nil) on `unlockedDisconnect`.

This approach seemed simpler than adding a message queue or a more sophisticated retry mechanism for failed sent messages.

### DIFF
--- a/messagix/client.go
+++ b/messagix/client.go
@@ -72,6 +72,9 @@ type Client struct {
 	unnecessaryCATRequests int
 
 	stopCurrentConnection atomic.Pointer[context.CancelFunc]
+
+	CanSendMessages  bool
+	SendMessagesCond *sync.Cond
 }
 
 func NewClient(cookies *cookies.Cookies, logger zerolog.Logger) *Client {
@@ -88,13 +91,15 @@ func NewClient(cookies *cookies.Cookies, logger zerolog.Logger) *Client {
 			},
 			Timeout: 60 * time.Second,
 		},
-		cookies:         cookies,
-		Logger:          logger,
-		lsRequests:      0,
-		graphQLRequests: 1,
-		platform:        cookies.Platform,
-		activeTasks:     make([]int, 0),
-		taskMutex:       &sync.Mutex{},
+		cookies:          cookies,
+		Logger:           logger,
+		lsRequests:       0,
+		graphQLRequests:  1,
+		platform:         cookies.Platform,
+		activeTasks:      make([]int, 0),
+		taskMutex:        &sync.Mutex{},
+		CanSendMessages:  false,
+		SendMessagesCond: sync.NewCond(&sync.Mutex{}),
 	}
 	cli.http.CheckRedirect = cli.checkHTTPRedirect
 
--- a/messagix/client.go
+++ b/messagix/client.go
@@ -225,6 +225,7 @@ func (c *Client) Connect() error {
 		connectionAttempts := 1
 		reconnectIn := 2 * time.Second
 		for {
+			c.disableSendingMessages() // In case we're reconnecting from a normal network error
 			connectStart := time.Now()
 			err := c.socket.Connect()
 			if ctx.Err() != nil {
@@ -401,19 +402,32 @@ func (c *Client) EnableSendingMessages() {
 	c.sendMessagesCond.L.Unlock()
 }
 
-func (c *Client) WaitUntilCanSendMessages(timeout time.Duration) error {
+func (c *Client) disableSendingMessages() {
 	c.sendMessagesCond.L.Lock()
-	defer c.sendMessagesCond.L.Unlock()
+	c.canSendMessages = false
+	c.sendMessagesCond.L.Unlock()
+}
 
+func (c *Client) WaitUntilCanSendMessages(timeout time.Duration) error {
 	timer := time.NewTimer(timeout)
 	defer timer.Stop()
+
+	done := make(chan struct{})
+	go func() {
+		c.sendMessagesCond.L.Lock()
+		defer c.sendMessagesCond.L.Unlock()
+		c.sendMessagesCond.Wait()
+		close(done)
+	}()
 
 	for !c.canSendMessages {
 		select {
 		case <-timer.C:
 			return fmt.Errorf("timeout waiting for canSendMessages")
-		default:
-			c.sendMessagesCond.Wait()
+		case <-done:
+			if c.canSendMessages {
+				return nil
+			}
 		}
 	}
 	return nil
--- a/messagix/client.go
+++ b/messagix/client.go
@@ -411,7 +411,7 @@ func (c *Client) WaitUntilCanSendMessages(timeout time.Duration) error {
 	for !c.canSendMessages {
 		select {
 		case <-timer.C:
-			return fmt.Errorf("timeout waiting for sending messages")
+			return fmt.Errorf("timeout waiting for canSendMessages")
 		default:
 			c.sendMessagesCond.Wait()
 		}
--- a/messagix/client.go
+++ b/messagix/client.go
@@ -73,8 +73,8 @@ type Client struct {
 
 	stopCurrentConnection atomic.Pointer[context.CancelFunc]
 
-	CanSendMessages  bool
-	SendMessagesCond *sync.Cond
+	canSendMessages  bool
+	sendMessagesCond *sync.Cond
 }
 
 func NewClient(cookies *cookies.Cookies, logger zerolog.Logger) *Client {
@@ -98,8 +98,8 @@ func NewClient(cookies *cookies.Cookies, logger zerolog.Logger) *Client {
 		platform:         cookies.Platform,
 		activeTasks:      make([]int, 0),
 		taskMutex:        &sync.Mutex{},
-		CanSendMessages:  false,
-		SendMessagesCond: sync.NewCond(&sync.Mutex{}),
+		canSendMessages:  false,
+		sendMessagesCond: sync.NewCond(&sync.Mutex{}),
 	}
 	cli.http.CheckRedirect = cli.checkHTTPRedirect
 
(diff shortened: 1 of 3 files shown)


messagix/events.go
```diff
@@ -109,6 +109,12 @@ func (s *Socket) handleReadyEvent(data *Event_Ready) error {
 	data.client = s.client
 	s.client.eventHandler(data.Finish())
 	s.previouslyConnected = true
+
+	s.client.SendMessagesCond.L.Lock()
+	s.client.CanSendMessages = true
+	s.client.SendMessagesCond.Signal()
```
tulir: Looking at the docs, this seems like it should be Broadcast, not Signal https://pkg.go.dev/sync#Cond.Broadcast

portal.go
```diff
@@ -630,6 +630,12 @@ func (portal *Portal) handleMatrixMessage(ctx context.Context, sender *User, evt
 	timings.convert = time.Since(start)
 	start = time.Now()
 
+	sender.Client.SendMessagesCond.L.Lock()
+	for !sender.Client.CanSendMessages {
+		sender.Client.SendMessagesCond.Wait()
+	}
+	sender.Client.SendMessagesCond.L.Unlock()
```
tulir: This should be in the `else` block that sends unencrypted messages. Also, I don't think this alone is enough, retries are necessary too (for example, what if it gets disconnected in the middle of sending?). Both the retries and this waiting should have an upper limit they'll try before giving up. The lock and bool probably shouldn't be exposed directly like this, it would be better to have a `WaitForConnection` function in `messagix.Client` that also takes a context or timeout

I'm thinking the end result should look something like this, replacing the current `ExecuteTasks` call below:

```go
retries := 0
for retries < 5 {
  sender.Client.WaitForConnection(15 * time.Second)
  resp, err = sender.Client.ExecuteTasks(tasks...)
  if err == nil {
    break
  }
  log.Err(err).Msg("...")
  retries++
  // Maybe also sleep here for a few seconds
}
```

javiercr: Great feedback, thank you! Done!

messagix/client.go
```diff
@@ -388,3 +393,28 @@ func (c *Client) GetTaskId() int {
...
+
+	timer := time.NewTimer(timeout)
+	defer timer.Stop()
+
+	for !c.canSendMessages {
+		select {
+		case <-timer.C:
+			return fmt.Errorf("timeout waiting for sending messages")
+		default:
+			c.sendMessagesCond.Wait()
```
tulir: I think this will still block until it's reconnected. Might need some ugly goroutines to actually give up when the timeout is reached

javiercr: This comment made realize that the error returned by `WaitUntilCanSendMessages` wasn't being properly handled. I've fixed that.

> I think this will still block until it's reconnected

However I can't see the scenario you're describing, could you elabore on that one? Thanks!

tulir: `Wait()` will block forever until `Broadcast()` is called. If calling `Broadcast()` is delayed because the connection keeps failing, then this can block for much longer than intended. However, I found this stackoverflow post pointing out it's possible to use `context.AfterFunc`: https://stackoverflow.com/a/77512028/2120293 (basically, you'd have `AfterFunc` call `Broadcast` to wake up all waiters and detect the context timeout)

Also just noticed that `canSendMessages` is never set to false (except when Client is initialized), and that WaitUntilCanSendMessages will deadlock if `canSendMessages` is `false` because it acquires the lock that EnableSendingMessages also wants.

messagix/client.go
```diff
@@ -388,3 +393,28 @@ func (c *Client) GetTaskId() int {
...
+		}
+	}
+	return nil
+}
```
javiercr: I think this would solve the two issues mentioned above (the blocking forever and the deadlock), wdyt @tulir?
```suggestion
func (c *Client) WaitUntilCanSendMessages(timeout time.Duration) error {
	timer := time.NewTimer(timeout)
	defer timer.Stop()

	done := make(chan struct{})
	go func() {
		c.sendMessagesCond.L.Lock()
		defer c.sendMessagesCond.L.Unlock()
		c.sendMessagesCond.Wait()
		close(done)
	}()

	for !c.canSendMessages {
		select {
		case <-timer.C:
			return fmt.Errorf("timeout waiting for canSendMessages")
		case <-done:
			if c.canSendMessages {
				return nil
			}
		}
	}
	return nil
}
```

messagix/client.go
```diff
@@ -388,3 +393,28 @@ func (c *Client) GetTaskId() int {
...
+	for !c.canSendMessages {
+		select {
+		case <-timer.C:
+			return fmt.Errorf("timeout waiting for canSendMessages")
+		default:
+			c.sendMessagesCond.Wait()
+		}
+	}
+	return nil
+}
```
javiercr: I think this would solve the two issues mentioned above (the blocking forever and the deadlock), wdyt @tulir?
```suggestion
func (c *Client) WaitUntilCanSendMessages(timeout time.Duration) error {
	timer := time.NewTimer(timeout)
	defer timer.Stop()

	done := make(chan struct{})
	go func() {
		c.sendMessagesCond.L.Lock()
		defer c.sendMessagesCond.L.Unlock()
		c.sendMessagesCond.Wait()
		close(done)
	}()

	for !c.canSendMessages {
		select {
		case <-timer.C:
			return fmt.Errorf("timeout waiting for canSendMessages")
		case <-done:
			if c.canSendMessages {
				return nil
			}
		}
	}
	return nil
}
```

messagix/client.go
javiercr: Good catch, indeed the timeout was not being triggered. I've proposed a solution for this in [the comment below](https://github.com/mautrix/meta/pull/63#discussion_r1607889584).

messagix/client.go
tulir: Hmm, I think it still has the deadlock issue: the lock is acquired during `Wait`, but the function doing `Broadcast` also wants the lock

Also, if the flag was theoretically set to false in between Wait returning and the `for` loop checking the condition, then it would get in a tight loop until the timer expires (because `done` would be closed, but the `canSendMessages` check would fail and not return). Right now that never happens because the flag is never set to false, but it should probably be set to false when disconnecting 🤔

This is getting so complicated that I'm thinking it might be better to just have retries with dumb exponential backoff in between (like max 5 attempts and sleep 2+4+8+16 seconds between attempts)

javiercr: > Hmm, I think it still has the deadlock issue: the lock is acquired during Wait, but the function doing Broadcast also wants the lock

From https://github.com/golang/go/blob/master/src/sync/cond.go#L66:

> Wait atomically unlocks c.L and suspends execution of the calling goroutine. After later resuming execution, Wait locks c.L before returning. Unlike in other systems, Wait cannot return unless awoken by Broadcast or Signal.

I think this prevent the deadlock you were describing, doesn't it?

> Right now that never happens because the flag is never set to false, but it should probably be set to false when disconnecting 🤔

Why should we do that? Client is always [disposed on User.unlockedDisconnect](https://github.com/mautrix/meta/blob/main/user.go#L1196).

tulir: > I think this prevent the deadlock you were describing, doesn't it?

Oh yes, so that part is fine

> Why should we do that? Client is always [disposed on User.unlockedDisconnect](https://github.com/mautrix/meta/blob/main/user.go#L1196).

Same as why #64 needs to cancel requests when disconnecting: `messagix.Client` handles normal network errors internally and reconnects automatically, such errors don't call `User.Disconnect`

javiercr: That makes sense!

Just added a `disableSendingMessages` that gets called every time we reconnect the socket: https://github.com/mautrix/meta/pull/63/commits/c93d8930eb4079e5495a733f2e02a8b5b395e45c

messagix/client.go
```diff
@@ -220,6 +225,7 @@ func (c *Client) Connect() error {
 		connectionAttempts := 1
 		reconnectIn := 2 * time.Second
 		for {
+			c.disableSendingMessages() // In case we're reconnecting from a normal network error
```
tulir: This should probably be after Connect returns, otherwise it stays enabled until the sleep at the end of the loop is done

tulir (approved):
Looks good other than the location of that disable call

messagix/client.go
javiercr: 🤔 Did you mean "stays disabled"? 

My idea for putting it here was that as soon as we go into this `for` loop, we know we need to connect / reconnect, and therefore there is no reason to have sending messages (`canSendMessages`) enabled (`true`). Therefore we better set it to false asap until the connection is completed, and the `EnableSendingMessages` have been called from `handleReadyEvent`.

Is there any other angle I'm not seeing? 

Thanks!

tulir: After a successful connection it's set to true. Later when the connection drops, `c.socket.Connect()` returns and the rest of the code in the loop is executed. There's a minimum of 2 seconds where the socket is disconnected, but `canSendMessages` is true, because the loop is still sleeping before a retry

javiercr: Gotcha, that makes sense. Thank you for the clarification. Just moved the `disableSendingMessages`.

Off topic, imho this code reads a bit weird

```golang
// client.go
func (c *Client) Connect() error {
	// ...
	go func() {
		// ...
		for {
			connectStart := time.Now()
			err := c.socket.Connect()
			c.disableSendingMessages()
			// ...
		}
	}()
	// ...
}
```

When I read this code for the first tiem, I had no idea that `c.socket.Connect()` would block in the socket's `readLoop`. Someone new to the code could be like, _why are we disabling sending messages right after connecting?_. From the outside, it's not obvious that the execution will only move forward once there is an error in the connection. I guess this is something we can improve in the future.

PR #64, opened 2024-05-19
Better errors for ResponseHandler and cancel ACK and response waits on disconnect

javiercr:
# Context

This PR continues the work on #63 (and should be merged after that one).

# Description

Given the following sequence diagram for sending a Message:

If a message is sent by the user, right an instant before a connection refresh start, the message will be delivered by Meta after receiving our `PUBLISH (/ls_req)` (1) packet.

However the connection refresh could be closing the socket before Meta's `PUBACK` (2), or the `PUBLISH(/ls_resp)` (3) are received by us.

In that scenario, the `makeLSRequest` will be locked until the timeout defined in `ResponseHandler.waitForDetails` is triggered.

This PR changes that behavior by exposing the `Context` used by `Client`, so that whenever that Context is cancelled (by calling `stopCurrentConnection`), we also notify `ResponseHandler.waitForDetails`. 

This PR also changes the signature for: `waitForPubACKDetails`,  `waitForSubACKDetails`, and `waitForPubResponseDetails` so that they return a different sentinel error depending on each scenario.

## TO BE DISCUSSED:

We need to discuss how to handle the new `ErrContextCancelled` in `Portal.handleMatrixMessage`. With the current implementation in this PR, the sent message will still be flagged as "Not delivered" in Beeper, despite having been delivered properly.

### DIFF
--- a/messagix/client.go
+++ b/messagix/client.go
@@ -72,6 +72,9 @@ type Client struct {
 	unnecessaryCATRequests int
 
 	stopCurrentConnection atomic.Pointer[context.CancelFunc]
+
+	CanSendMessages  bool
+	SendMessagesCond *sync.Cond
 }
 
 func NewClient(cookies *cookies.Cookies, logger zerolog.Logger) *Client {
@@ -88,13 +91,15 @@ func NewClient(cookies *cookies.Cookies, logger zerolog.Logger) *Client {
 			},
 			Timeout: 60 * time.Second,
 		},
-		cookies:         cookies,
-		Logger:          logger,
-		lsRequests:      0,
-		graphQLRequests: 1,
-		platform:        cookies.Platform,
-		activeTasks:     make([]int, 0),
-		taskMutex:       &sync.Mutex{},
+		cookies:          cookies,
+		Logger:           logger,
+		lsRequests:       0,
+		graphQLRequests:  1,
+		platform:         cookies.Platform,
+		activeTasks:      make([]int, 0),
+		taskMutex:        &sync.Mutex{},
+		CanSendMessages:  false,
+		SendMessagesCond: sync.NewCond(&sync.Mutex{}),
 	}
 	cli.http.CheckRedirect = cli.checkHTTPRedirect
 
--- a/messagix/client.go
+++ b/messagix/client.go
@@ -67,6 +67,7 @@ type Client struct {
 	endpoints       map[string]string
 	taskMutex       *sync.Mutex
 	activeTasks     []int
+	ctx             context.Context
 
 	catRefreshLock         sync.Mutex
 	unnecessaryCATRequests int
@@ -217,6 +218,7 @@ func (c *Client) Connect() error {
 		return err
 	}
 	ctx, cancel := context.WithCancel(context.TODO())
+	c.ctx = ctx
 	oldCancel := c.stopCurrentConnection.Swap(&cancel)
 	if oldCancel != nil {
 		(*oldCancel)()
@@ -392,4 +394,8 @@ func (c *Client) GetTaskId() int {
 
 	c.activeTasks = append(c.activeTasks, id)
 	return id
+}
+
+func (c *Client) Context() context.Context {
+	return c.ctx
 }
--- a/messagix/client.go
+++ b/messagix/client.go
@@ -67,7 +67,6 @@ type Client struct {
 	endpoints       map[string]string
 	taskMutex       *sync.Mutex
 	activeTasks     []int
-	ctx             context.Context
 
 	catRefreshLock         sync.Mutex
 	unnecessaryCATRequests int
@@ -218,7 +217,6 @@ func (c *Client) Connect() error {
 		return err
 	}
 	ctx, cancel := context.WithCancel(context.TODO())
-	c.ctx = ctx
 	oldCancel := c.stopCurrentConnection.Swap(&cancel)
 	if oldCancel != nil {
 		(*oldCancel)()
--- a/messagix/client.go
+++ b/messagix/client.go
@@ -395,7 +395,3 @@ func (c *Client) GetTaskId() int {
 	c.activeTasks = append(c.activeTasks, id)
 	return id
 }
-
-func (c *Client) Context() context.Context {
-	return c.ctx
-}
--- a/messagix/client.go
+++ b/messagix/client.go
@@ -263,7 +263,6 @@ func (c *Client) Disconnect() {
 		(*fn)()
 	}
 	c.socket.Disconnect()
-	c.socket.conn = nil
 }
 
 func (c *Client) SaveSession(path string) error {
--- a/messagix/client.go
+++ b/messagix/client.go
@@ -73,8 +73,8 @@ type Client struct {
 
 	stopCurrentConnection atomic.Pointer[context.CancelFunc]
 
-	CanSendMessages  bool
-	SendMessagesCond *sync.Cond
+	canSendMessages  bool
+	sendMessagesCond *sync.Cond
 }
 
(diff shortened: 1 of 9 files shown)


messagix/responsehandler.go
```diff
@@ -19,6 +20,10 @@ type ResponseHandler struct {
 	packetChannels  map[uint16]chan interface{}
 }
 
+var ErrTimeout = fmt.Errorf("timeout waiting for response")
+var ErrContextCancelled = fmt.Errorf("context cancelled before response received")
+var ErrChannelNotFound = fmt.Errorf("channel not found for packet")
```
javiercr, line 25: Happy to take any suggestions for better error messages here.

messagix/threads.go
```diff
@@ -21,7 +21,7 @@ func (c *Client) ExecuteTasks(tasks ...socket.Task) (*table.LSTable, error) {
 
 	resp, err := c.socket.makeLSRequest(payload, 3)
 	if err != nil {
-		return nil, fmt.Errorf("failed to send LS request: %v", err)
+		return nil, err
```
javiercr, line 24: Heads up: this is gonna cause changes in the log files / records for users of the bridge, including Beeper infra. As the error strings will be different now.

messagix/responsehandler.go
```diff
@@ -79,20 +96,22 @@ func castIfNotNil[T any](i interface{}) *T {
...
+	case <-p.client.Context().Done():
+		p.deleteDetails(packetId, channelType)
+		return nil, ErrContextCancelled
 	case response := <-ch:
 		p.deleteDetails(packetId, channelType)
-		return response
+		return response, nil
 	case <-time.After(packetTimeout):
 		p.deleteDetails(packetId, channelType)
-		// TODO this should probably be an error
```
javiercr, line 94: I guess the TODO comment here was right :)

messagix/socket.go
```diff
@@ -366,7 +369,11 @@ func (s *Socket) sendPublishPacket(topic Topic, jsonData string, packet *packets
...
 	}
-	ack := s.responseHandler.waitForPubACKDetails(packetId)
+	ack, err := s.responseHandler.waitForPubACKDetails(packetId)
+	if err != nil {
+		s.responseHandler.deleteDetails(packetId, RequestChannel)
+		return packetId, err
+	}
 	if ack == nil {
 		s.responseHandler.deleteDetails(packetId, RequestChannel)
 		return packetId, fmt.Errorf("puback timeout")
```
tulir: This part should never happen now that it's an error

messagix/socket.go
```diff
@@ -400,11 +407,8 @@ func (s *Socket) makeLSRequest(payload []byte, t int) (*Event_PublishResponse, e
 		return nil, err
 	}
 
-	resp := s.responseHandler.waitForPubResponseDetails(packetId)
-	if resp == nil {
-		return nil, fmt.Errorf("publish response timeout")
-	}
-	return resp, nil
+	resp, err := s.responseHandler.waitForPubResponseDetails(packetId)
+	return resp, err
```
tulir:
```suggestion
	return s.responseHandler.waitForPubResponseDetails(packetId)
```

messagix/responsehandler.go
```diff
@@ -79,20 +96,22 @@ func castIfNotNil[T any](i interface{}) *T {
...
-func (p *ResponseHandler) waitForDetails(packetId uint16, channelType ChannelType) interface{} {
+func (p *ResponseHandler) waitForDetails(packetId uint16, channelType ChannelType) (interface{}, error) {
 	ch, ok := p.getChannel(packetId, channelType)
 	if !ok {
-		return nil
+		return nil, ErrChannelNotFound
 	}
 
 	select {
+	case <-p.client.Context().Done():
```
tulir: This is in the same package so it can just access the field directly:

```suggestion
	case <-p.client.ctx.Done():
```

messagix/client.go
```diff
@@ -393,3 +395,7 @@ func (c *Client) GetTaskId() int {
 	c.activeTasks = append(c.activeTasks, id)
 	return id
 }
+
+func (c *Client) Context() context.Context {
+	return c.ctx
+}
```
tulir: This is probably not necessary, I don't think anything outside the package should use the context

```suggestion
```

messagix/socket.go
javiercr: good catch, done!

messagix/responsehandler.go
tulir: Could say "client disconnected" or something instead of context cancelled. The two others are probably fine

messagix/client.go
```diff
@@ -217,6 +218,7 @@ func (c *Client) Connect() error {
 		return err
 	}
 	ctx, cancel := context.WithCancel(context.TODO())
+	c.ctx = ctx
 	oldCancel := c.stopCurrentConnection.Swap(&cancel)
```
tulir: I just realized that this context is not cancelled until a new `Connect` call, and that it's not actually per socket connection because this function has a loop that retries socket connections forever (until stopped)

It might actually be better to not mess with contexts at all and instead just close all channels when socket.Connect returns. Basically add a `cancelAllRequests` to `ResponseHandler` which acquires the lock, loops through every request and packet channel and closes them. The places that consume the channels will then get a `nil`, which can be interpreted as the request being cancelled due to disconnection

messagix/responsehandler.go
tulir: Might not have context cancellations at all if you use the different approach I mentioned

messagix/client.go
javiercr: > I just realized that this context is not cancelled until a new Connect call

The context is also cancelled here: 
```go
func (c *Client) Disconnect() {
	if fn := c.stopCurrentConnection.Load(); fn != nil {
		(*fn)()
	}
	c.socket.Disconnect()
}
```

which is called from `User.unlockedDisconnect()` and therefor from `User.FullReconnect()`. 

> and that it's not actually per socket connection because this function has a loop that retries socket connections forever (until stopped)

what do you mean here 🤔? There is only one `Socket` per `Client`, and therefore only one `ResponseHandler`, isn't that right?

tulir: Ah yes missed that. Anyway, `Client.Connect` calls `socket.Connect` forever until something tells the client to disconnect (see the loop inside the goroutine a few lines below this). So this will cancel requests during a full reconnect, but not during a reconnect caused by a network error.

messagix/responsehandler.go
```diff
@@ -103,9 +103,6 @@ func (p *ResponseHandler) waitForDetails(packetId uint16, channelType ChannelTyp
 	}
 
 	select {
-	case <-p.client.ctx.Done():
-		p.deleteDetails(packetId, channelType)
-		return nil, ErrContextCancelled
 	case response := <-ch:
 		p.deleteDetails(packetId, channelType)
 		return response, nil
```
tulir: Canceling looks good now, just need to handle the `response` being nil here and return an error instead (when a channel is closed, it'll wake up selectors, but will give nil as the value).

```suggestion
	case response := <-ch:
		if response == nil {
			return nil, ErrConnectionClosed
		}
		p.deleteDetails(packetId, channelType)
		return response, nil
```

(ErrConnectionClosed needs to be defined with the errors above)

javiercr: > Basically add a cancelAllRequests to ResponseHandler which acquires the lock, loops through every request and packet channel and closes them. The places that consume the channels will then get a nil, which can be interpreted as the request being cancelled due to disconnection

Done.

javiercr: Thanks, done!

tulir (approved):
Looks like this should work now

PR #68, opened 2024-05-30
Fixes WhatsApp socket (e2e chats) reconnection on 401

javiercr:
When a user [log outs the device from the Messenger mobile app](https://www.facebook.com/help/messenger-app/986015482240175/?cms_platform=android-app&helpref=platform_switcher) for e2e chats, whatsmeow dispatches a `events.LoggedOut` event rather than a `events.ConnectFailure`:

```go
case code == "401" && conflictType == "device_removed":
	cli.expectDisconnect()
	cli.Log.Infof("Got device removed stream error, sending LoggedOut event and deleting session")
	go cli.dispatchEvent(&events.LoggedOut{OnConnect: false, Reason: events.ConnectFailureLoggedOut})
        //...
```
https://github.com/tulir/whatsmeow/blob/main/connectionevents.go#L34-L37

This PR fixes that, so even if the device is removed for e2e chats, we keep the bridge connected to the WA socket.

### DIFF
--- a/user.go
+++ b/user.go
@@ -1040,8 +1040,9 @@ func (user *User) e2eeEventHandler(rawEvt any) {
 		user.BridgeState.Send(user.waState)
 		go user.sendMarkdownBridgeAlert(context.TODO(), "Error in WhatsApp connection: %s", evt.PermanentDisconnectDescription())
 	case events.PermanentDisconnect:
-		cf, ok := evt.(*events.ConnectFailure)
+		cf, ok := evt.(*events.LoggedOut)
 		if ok && cf.Reason == events.ConnectFailureLoggedOut && user.canReconnect() {
+			user.WADevice = nil
 			user.log.Debug().Msg("Doing full reconnect after WhatsApp 401 error")
 			go user.FullReconnect()
 		}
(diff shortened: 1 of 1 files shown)


user.go
```diff
@@ -1040,8 +1040,9 @@ func (user *User) e2eeEventHandler(rawEvt any) {
 		user.BridgeState.Send(user.waState)
 		go user.sendMarkdownBridgeAlert(context.TODO(), "Error in WhatsApp connection: %s", evt.PermanentDisconnectDescription())
 	case events.PermanentDisconnect:
-		cf, ok := evt.(*events.ConnectFailure)
-		if ok && cf.Reason == events.ConnectFailureLoggedOut && user.canReconnect() {
+		cf, ok := evt.(*events.LoggedOut)
+		if ok && cf.Reason == events.ConnectFailureLoggedOut && !cf.OnConnect && user.canReconnect() {
```
javiercr, line 1044: The `!cf.OnConnect` clause avoids reconnecting if the 401 comes from the user logging out from facebook totally (not just the e2e/WA socket log out described in the PR).

In that scenario `messagix.Event_PermanentError` is also triggered.

tulir (approved):
Shouldn't break anything at least, not sure if it actually changes anything in the common cases though

PR #70, opened 2024-06-06
Reconnect WhatsApp socket (e2e messaging) on 415 `ConnectFailureNotFound`

javiercr:
We don't know the reasons for 415 errors, but resetting the WA device seems to help.

### DIFF
--- a/user.go
+++ b/user.go
@@ -1040,12 +1040,19 @@ func (user *User) e2eeEventHandler(rawEvt any) {
 		user.BridgeState.Send(user.waState)
 		go user.sendMarkdownBridgeAlert(context.TODO(), "Error in WhatsApp connection: %s", evt.PermanentDisconnectDescription())
 	case events.PermanentDisconnect:
-		cf, ok := evt.(*events.LoggedOut)
-		if ok && cf.Reason == events.ConnectFailureLoggedOut && !cf.OnConnect && user.canReconnect() {
-			user.WADevice = nil
-			user.log.Debug().Msg("Doing full reconnect after WhatsApp 401 error")
-			go user.FullReconnect()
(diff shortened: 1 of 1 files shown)


user.go
```diff
@@ -1058,6 +1065,12 @@ func (user *User) e2eeEventHandler(rawEvt any) {
 	}
 }
 
+func (user *User) resetWADeviceAndReconnect(reason events.ConnectFailureReason) {
+	user.log.Debug().Int("ConnectFailureReason", int(reason)).Msg("Resetting WADevice and reconnecting")
+	user.WADevice = nil
+	go user.FullReconnect()
```
tulir: There isn't really any need for a full reconnect in the 415 case, just

```go
if user.E2EEClient != nil {
  user.E2EEClient.Disconnect()
  user.E2EEClient = nil
}
err := user.connectE2EE()
```

should be enough

(the 401 case probably does need a full reconnect though to check if the user got completely logged out or not)

user.go
```diff
@@ -1058,6 +1065,12 @@ func (user *User) e2eeEventHandler(rawEvt any) {
 	}
 }
 
+func (user *User) resetWADeviceAndReconnect(reason events.ConnectFailureReason) {
+	user.log.Debug().Int("ConnectFailureReason", int(reason)).Msg("Resetting WADevice and reconnecting")
+	user.WADevice = nil
```
tulir, line 1076: I don't think this is enough for 415s, `WADeviceID` also needs to be cleared (otherwise `connectE2EE` will just reload the existing device from the store) and it would probably be a good idea to delete the device completely from the db.

For 401s, whatsmeow will delete the device from the db, so just clearing the cache on the bridge side is enough

PR #73, opened 2024-06-28
add registration for push notifications

the-newman:


tulir (approved):
Looks fairly sensible

the-newman:
Thanks for the review @tulir

PR #77, opened 2024-06-29
PLAT-25704: Add megabridge login flows 

JJTech0130:
Completes the login flows for both Facebook and Instagram. Also fixes some bugs left behind from the last PR.
Note that it doesn't save the results of the login anywhere, will leave that for the next PR.

Note to reviewer: I still need help with the whole `example-config.yml` thing, it appears to override whatever is in config.yaml no matter what I do? In any case doesn't really affect anything right now.

Also closes PLAT-25705 because it implements proper selection between facebook and instagram bridge modes, and allows using both at once.


pkg/connector/login.go
```diff
@@ -5,43 +5,151 @@ import (
...
 	"maunium.net/go/mautrix/bridgev2"
+
+	"go.mau.fi/mautrix-meta/messagix"
+	"go.mau.fi/mautrix-meta/messagix/cookies"
+	"go.mau.fi/mautrix-meta/messagix/table"
+	"go.mau.fi/mautrix-meta/messagix/types"
 )
 
+var FACEBOOK_COOKIES_FLOW = "cookies-facebook"
+var INSTAGRAM_COOKIES_FLOW = "cookies-instagram"
```
tulir: `const FlowIDFacebookCookies` would be better

pkg/connector/login.go
```diff
@@ -5,43 +5,151 @@ import (
...
 
-var _ bridgev2.LoginProcessCookies = (*CookieLogin)(nil)
+func requiredCookiesToKeys(requiredCookies []cookies.MetaCookieName) []string {
+	var keys []string
+	// Turn the array of MetaCookieName into an array of strings
+	for _, key := range requiredCookies {
+		keys = append(keys, string(key))
+	}
+	return keys
+}
```
tulir: I added a generic utility for this: `exslices.CastToString` should work with any string type aliases https://github.com/mautrix/go-util/commit/4fa40bf64652a2ac475e57b456cbfef4fb7ee1cf

pkg/connector/login.go
```diff
@@ -5,43 +5,151 @@ import (
...
+		Name:        "Facebook Cookies",
+		Description: "Login using cookies from Facebook Messenger",
+		ID:          FACEBOOK_COOKIES_FLOW,
+	}
+	instagram := bridgev2.LoginFlow{
+		Name:        "Instagram Cookies",
+		Description: "Login using cookies from Instagram",
+		ID:          INSTAGRAM_COOKIES_FLOW,
+	}
+	if m.Config == nil {
```
tulir:
```suggestion
	if m.Config == nil || m.Config.Mode == "" {
```

go.mod
```diff
@@ -26,14 +26,14 @@ require (
 require (
 	filippo.io/edwards25519 v1.1.0 // indirect
 	github.com/coreos/go-systemd/v22 v22.5.0 // indirect
-	github.com/gorilla/mux v1.8.0 // indirect
-	github.com/mattn/go-isatty v0.0.19 // indirect
+	github.com/gorilla/mux v1.8.1 // indirect
```
tulir: gorilla packages should not be updated

JJTech0130:
I rolled back the gorilla update (and the other spurious package updates), unsure why the `go vet` check is now failing? I've run `go vet` and `go mod tidy` locally and they both exit cleanly with no modifications.

PR #78, opened 2024-07-01
PLAT-25752: Add Megabridge session store

JJTech0130:
Now login sessions are actually stored, rather than just being thrown away. Yay!


pkg/store/session_store.go
```diff
@@ -0,0 +1,78 @@
...
+	var metaID sql.NullInt64
+	var waDeviceID sql.NullInt32
+	var platform sql.NullString
+	var newCookies cookies.Cookies
+
+	err := row.Scan(
+		&metaID,
+		&platform,
+		&waDeviceID,
+		&dbutil.JSON{Data: &newCookies},
```
JJTech0130: How exactly does this automatic JSON storage work? Will it store the platform and, more importantly the push keys?

pkg/store/session_store.go
```diff
@@ -0,0 +1,78 @@
...
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	if platform.String == "instagram" {
```
JJTech0130: Should just store them using their number directly in the DB

pkg/store/session_store.go
```diff
@@ -0,0 +1,78 @@
...
+
+type MetaSessionQuery struct {
+	*dbutil.QueryHelper[*MetaSession]
+}
+
+func (q *MetaSessionQuery) GetByMetaID(ctx context.Context, metaID int64) (*MetaSession, error) {
+	return q.QueryOne(ctx, getUserByMetaIDQuery, metaID)
+}
+
+func (q *MetaSessionQuery) Insert(ctx context.Context, session *MetaSession) error {
```
JJTech0130: Probably want some kind of "update" function, a way of keeping in-sync as changes are made: I assume the cookies need refreshing?

pkg/connector/login.go
```diff
@@ -137,6 +141,70 @@ func (m *MetaCookieLogin) SubmitCookies(ctx context.Context, strCookies map[stri
 
 	FBID := m.getFBID(currentUser, initialTable)
```
JJTech0130: This *probably* shouldn't be all in `SubmitCookies`

pkg/connector/login.go
```diff
@@ -137,6 +141,70 @@ func (m *MetaCookieLogin) SubmitCookies(ctx context.Context, strCookies map[stri
...
 
+	// Store the cookies in the database
+
+	q := m.Main.store.GetSessionQuery()
+	session, err := q.GetByMetaID(ctx, FBID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get existing session: %w", err)
+	}
+	if session != nil {
+		log.Debug().Int64("fbid", FBID).Msg("Deleting existing session")
```
JJTech0130: Are there ever going to be multiple open sessions for the same FBID? Would only happen if sharing a Facebook/Instagram account, which would be weird so I guess not

pkg/connector/login.go
```diff
@@ -137,6 +141,70 @@ func (m *MetaCookieLogin) SubmitCookies(ctx context.Context, strCookies map[stri
...
+		Cookies: c,
+	}
+	err = q.Insert(ctx, &newSession)
+	if err != nil {
+		return nil, fmt.Errorf("failed to save new session: %w", err)
+	}
+
+	// Store the association between the Matrix user and the cookies
+
+	login_id := networkid.UserLoginID(fmt.Sprint(FBID))
```
JJTech0130: break this out

pkg/connector/login.go
```diff
@@ -137,6 +141,70 @@ func (m *MetaCookieLogin) SubmitCookies(ctx context.Context, strCookies map[stri
...
+
+	ul, err := m.Main.Bridge.GetExistingUserLoginByID(ctx, login_id)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get existing login: %w", err)
+	}
+	if ul != nil && ul.UserMXID != m.User.MXID {
```
JJTech0130, line 131: Why is this logic not part of megabridge itself, tbh?

pkg/connector/login.go
```diff
@@ -137,6 +141,70 @@ func (m *MetaCookieLogin) SubmitCookies(ctx context.Context, strCookies map[stri
...
+		ul.Delete(ctx, status.BridgeState{StateEvent: status.StateLoggedOut, Error: "overridden-by-another-user"}, false)
+		ul = nil
+	}
+	if ul == nil {
+		ul, err = m.User.NewLogin(ctx, &database.UserLogin{
+			ID: login_id,
+			Metadata: database.UserLoginMetadata{
+				StandardUserLoginMetadata: database.StandardUserLoginMetadata{
+					// TODO: Is this the correct metadata? Any extra?
+					RemoteName: currentUser.GetName(),
```
JJTech0130: do we need to store this? What is the point of this login metadata, where is it used?

pkg/connector/client.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"go.mau.fi/mautrix-meta/messagix"
+	//"go.mau.fi/mautrix-meta/pkg/store"
+	"maunium.net/go/mautrix/bridgev2"
+	"maunium.net/go/mautrix/bridgev2/networkid"
+)
+
+type MetaClient struct {
+	Main *MetaConnector
+	//session *store.MetaSession
```
JJTech0130: Should we store the Session just for easier database updates? Could make the Session wrap Client instead of Cookies... might be better anyway, hmm...

JJTech0130 (commented):
Just a bunch of notes for myself tomorrow

JJTech0130:
A bunch of changes to the way the session is stored ended up being made in #79

PR #79, opened 2024-07-03
PLAT-25728: Implement Client interface

JJTech0130:
Still a work in progress, stacked on #78

PR #81, opened 2024-07-05
PLAT-25729 Receive messages

JJTech0130:
Stacked on #79 

Closes PLAT-25729, PLAT-25808, and PLAT-25844


JJTech0130:
Just realized I forgot to push all of these commits...

pkg/connector/client.go
```diff
@@ -140,11 +286,11 @@ func (m *MetaClient) Disconnect() {
...
 // GetCapabilities implements bridgev2.NetworkAPI.
 func (m *MetaClient) GetCapabilities(ctx context.Context, portal *bridgev2.Portal) *bridgev2.NetworkRoomCapabilities {
-	panic("unimplemented")
+	return &bridgev2.NetworkRoomCapabilities{}
 }
 
 // GetChatInfo implements bridgev2.NetworkAPI.
-func (m *MetaClient) GetChatInfo(ctx context.Context, portal *bridgev2.Portal) (*bridgev2.PortalInfo, error) {
+func (m *MetaClient) GetChatInfo(ctx context.Context, portal *bridgev2.Portal) (*bridgev2.ChatInfo, error) {
 	panic("unimplemented")
```
JJTech0130: @tulir 
Still need to figure out what exactly to do about `GetChatInfo` -- in previous commits, I had a version where I stored all thread metadata separately in-memory (doesn't need to be written to DB because we receive it all at startup in the initial table), but that is obviously suboptimal.

I don't think there is a proper way to accomplish it without the `CreateMatrixRoom` hack I have right now, because otherwise there is nowhere to store thread metadata before the room is created. And the hack doesn't work if the user deletes and recreates the portal without restarting.

(This does not apply to `GetUserInfo` and ghost metadata as far as I can tell, because that can just be set at any time, even before the ghost has joined any rooms)

pkg/connector/client.go
```diff
@@ -2,27 +2,46 @@ package connector
...
 	"maunium.net/go/mautrix/bridgev2"
+	"maunium.net/go/mautrix/bridgev2/database"
 	"maunium.net/go/mautrix/bridgev2/networkid"
+	"maunium.net/go/mautrix/event"
+
+	metaTypes "go.mau.fi/mautrix-meta/messagix/types"
 )
 
+type metaEvent struct {
+	context context.Context
```
JJTech0130, line 31: Oh, another thing I wanted to ask about: I store a Context in this struct so that it can be passed into the `incomingEvents` channel, which is picked up by `go handleMetaEventLoop()` -- 

I thought a Context would be desirable for cancellation etc. in the future, especially since it is a goroutine, but I realize that you're not supposed to put Contexts in a struct... is there a better solution in this case?

pkg/connector/client.go
JJTech0130: Came up with a solution that works, thought it might be a bit wasteful (depends on if you prioritize memory possibly used by un-bridged chats or network requests):
Just re-requesting the entire initial table and manually parsing out the metadata right there.

PR #85, opened 2024-07-19
PLAT-25917 Catch and handle invalid access token response to upload requests

Fizzadar:


messagix/mercury.go
```diff
@@ -53,6 +54,9 @@ func (c *Client) SendMercuryUploadRequest(ctx context.Context, threadID int64, m
 
 	_, respBody, err := c.MakeRequest(url, "POST", h, payload, types.NONE)
 	if err != nil {
+		if strings.Contains(err.Error(), "access token is no longer valid") {
+			return nil, ErrTokenInvalidated
+		}
```
tulir: This is just checking if it's the error and returning the same error 🤔

messagix/mercury.go
```diff
@@ -53,6 +54,9 @@ func (c *Client) SendMercuryUploadRequest(ctx context.Context, threadID int64, m
...
 		return nil, fmt.Errorf("failed to send MercuryUploadRequest: %v", err)
```
tulir: this is the actual fix:

```suggestion
		return nil, fmt.Errorf("failed to send MercuryUploadRequest: %w", err)
```

tulir (changes requested):

messagix/mercury.go
tulir:
```suggestion
```

Fizzadar: 🤦

tulir (approved):

messagix/http.go
```diff
@@ -190,7 +190,7 @@ func (c *Client) makeRequestDirect(url string, method string, headers http.Heade
 		}
 	}()
 	if err != nil {
-		c.UpdateProxy(fmt.Sprintf("http request error: %v", err.Error()))
+		c.UpdateProxy(fmt.Sprintf("http request error: %w", err.Error()))
```
tulir: This one is still wrong:

```suggestion
		c.UpdateProxy(fmt.Sprintf("http request error: %s", err.Error()))
```

PR #89, opened 2024-07-24
PLAT-25984 Searching, identifier lookups

JJTech0130:
The first commit got directly pushed to the `megabridge` branch, but here is a follow up that actually implements searching as a separate thing. ~~Untested as Tulir has not yet implemented the command to trigger it.~~
Tested now


JJTech0130:
This should now resolve PLAT-25948 as well

PR #90, opened 2024-07-25
Retry mercury upload requests when we fail to parse the response

Fizzadar:


tulir (approved):
Code looks good, although it would be nice to confirm whether retries actually help

Fizzadar:
> Code looks good, although it would be nice to confirm whether retries actually help

No forthcoming repros unfortunately, going to merge, review and revert if it make no difference.

PR #93, opened 2024-07-31
PLAT-25849 Message edits

JJTech0130:
Things with `msgconv` are starting to get a little bit hairy, should probably rewrite that whole thing...
In any case, would really appreciate a review @tulir, some of that edit part stuff I think is not quite right.

### DIFF
--- a/pkg/connector/client.go
+++ b/pkg/connector/client.go
@@ -413,8 +413,8 @@ func (m *MetaClient) handleTable(ctx context.Context, tbl *table.LSTable) {
 				return &bridgev2.ConvertedEdit{
 					ModifiedParts: []*bridgev2.ConvertedEditPart{
 						{
-							Part: textPart,
-							Type: event.EventMessage,
+							Part:    textPart,
+							Type:    event.EventMessage,
 							Content: m.messageConverter.MetaToMatrixText(ctx, data.Text, nil, portal),
 						},
 					},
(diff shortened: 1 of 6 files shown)


pkg/connector/client.go
```diff
@@ -383,10 +381,53 @@ func (m *MetaClient) handleTable(ctx context.Context, tbl *table.LSTable) {
...
+					ModifiedParts: []*bridgev2.ConvertedEditPart{
+						{
+							Part: textPart,
+							Type: event.EventMessage,
+							Content: &event.MessageEventContent{
+								MsgType:       event.MsgText,
+								Body:          data.Text,
+								Format:        event.FormatHTML,
+								FormattedBody: m.messageConverter.MetaToMatrixText(ctx, data.Text, nil, portal).Body,
+							},
```
tulir:
```suggestion
							Content: m.messageConverter.MetaToMatrixText(ctx, data.Text, nil, portal),
```

pkg/connector/msgconv/mentions.go
```diff
@@ -51,15 +51,21 @@ var (
...
-		zerolog.Ctx(ctx).Err(err).Msg("Failed to parse mentions")
+
+	var mentions *socket.Mentions
+
+	if rawMentions != nil {
+		mentionsP, err := rawMentions.Parse()
+		if err != nil {
+			zerolog.Ctx(ctx).Err(err).Msg("Failed to parse mentions")
+		}
+		mentions = &mentionsP
```
tulir: It would be easier to change `MentionData.Parse` to return nil if the input is nil than add a whole if like this

PR #100, opened 2024-09-02
Make config display name template instance public

Fizzadar:
Allows creating a config instance directly without YAML.


tulir:
Hmm, I think the signal connector's style is nicer where it decodes an empty string as yaml in order to call the decode function (`cfg.Network.Decode(signalConnector.Config)`)

Fizzadar:
Aha yes this is much nicer!

PR #103, opened 2024-09-19
metaid: use int64 for 32-bit compatibility

NicolasDerumigny:
Fixes a resurgence of #5 following ec80bfbf.

### DIFF
--- a/pkg/metaid/ids.go
+++ b/pkg/metaid/ids.go
@@ -10,19 +10,19 @@ import (
 )
 
 func MakeWAUserID(jid types.JID) networkid.UserID {
-	return networkid.UserID(strconv.Itoa(int(jid.UserInt())))
+	return networkid.UserID(strconv.FormatUint(jid.UserInt(), 10))
 }
 
 func MakeUserID(user int64) networkid.UserID {
-	return networkid.UserID(strconv.Itoa(int(user)))
+	return networkid.UserID(strconv.FormatInt(user, 10))
 }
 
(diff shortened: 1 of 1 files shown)


pkg/metaid/ids.go
```diff
@@ -34,21 +34,21 @@ func MakeWAPortalID(jid types.JID) networkid.PortalID {
...
 
 func MakeFBPortalID(portal int64) networkid.PortalID {
-	return networkid.PortalID(strconv.Itoa(int(portal)))
+	return networkid.PortalID(strconv.FormatInt(portal, 10))
 }
 
 func ParseUserID(user networkid.UserID) int64 {
-	i, _ := strconv.Atoi(string(user))
+	i, _ := strconv.ParseInt(string(user), 10, 64)
 	return int64(i)
```
tulir: These casts are redundant now since `i` is already `int64` (also `ParseIDFromString`, `ParseUserLoginID` and `ParseFBPortalID`)

NicolasDerumigny: I do not think so. Atoi returns a `int` (https://pkg.go.dev/strconv#Atoi) which is an `int32` on 32-bit arch, which is then converted to an `int64` in the return cast, but it is too late.

tulir: I mean the casts, not the parsing

NicolasDerumigny: Ow, my bad! Thanks, I'll correct that

tulir (approved):

PR #115, opened 2025-02-09
legacyprovision: fix RespError json serialization

rnons:
A POC snippet
```go
package main

import (
	"encoding/json"
	"log"
	"net/http"

	"go.mau.fi/mautrix-meta/pkg/connector"
)

func jsonResponse(w http.ResponseWriter, status int, response any) {
	w.Header().Add("Content-Type", "application/json")
	w.WriteHeader(status)
	_ = json.NewEncoder(w).Encode(response)
}

func main() {
	err := connector.ErrLoginChallenge
	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		// Returns {"ErrCode": "...", "Err": "..."}
		jsonResponse(w, err.StatusCode, err)
		// Returns {"errcode": "...", "err": "..."}
		// err.Write(w)
	})

	log.Fatal(http.ListenAndServe(":8080", nil))
}
```

Seems `RespError` becomes `any` in `jsonResponse`, so `func (e *RespError) MarshalJSON() ([]byte, error)` is not used.

### DIFF
--- a/cmd/mautrix-meta/legacyprovision.go
+++ b/cmd/mautrix-meta/legacyprovision.go
@@ -67,7 +67,7 @@ func legacyProvLogin(w http.ResponseWriter, r *http.Request) {
 		zerolog.Ctx(ctx).Err(err).Msg("Failed to log in")
 		var respErr bridgev2.RespError
 		if errors.As(err, &respErr) {
-			jsonResponse(w, respErr.StatusCode, &respErr)
+			respErr.Write(w)
 		} else {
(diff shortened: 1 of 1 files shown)


sumnerevans (approved):

tulir (approved):
Becoming `any` doesn't prevent methods from being called, the actual issue is that `bridgev2.RespError` doesn't have MarshalJSON at all. The fix is good either way though

PR #121, opened 2025-02-20
Added temporary support for Dendrite instances

mrupnikm:
I added a simple way to add extra command to ignore unsupported server so it can be compatible with [Dendrite instances](https://github.com/element-hq/dendrite). Also changed bash to sh for ease of compiling on multiple distros.

The bellow example worked for me:

```
image: ...
command: /docker-run.sh --ignore-unsupported-server
volumes: ...
```


mrupnikm:
@tulir do you think this is possible to merge?

mrupnikm:
Anything on this simple change?

PR #124, opened 2025-03-17
capabilities: separate capabilities for e2e chat

rnons:


pkg/connector/capabilities.go
```diff
@@ -149,13 +150,21 @@ var metaCaps = &event.RoomFeatures{
...
 	metaCapsWithThreads = ptr.Clone(metaCaps)
 	metaCapsWithThreads.ID += "+communitygroup"
 	metaCapsWithThreads.Thread = event.CapLevelFullySupported
 
+	metaCapsWithE2E = ptr.Clone(metaCaps)
+	metaCapsWithE2E.ID += "+e2e"
+	metaCapsWithE2E.File[event.MsgVideo].MaxSize = MaxFileSizeWithE2E
+	metaCapsWithE2E.File[event.MsgAudio].MaxSize = MaxFileSizeWithE2E
+	metaCapsWithE2E.File[event.MsgFile].MaxSize = MaxFileSizeWithE2E
+	metaCapsWithE2E.File[event.CapMsgVoice].MaxSize = MaxFileSizeWithE2E
```
tulir: This needs to clone the map and items inside like igCaps does below, otherwise it mutates the original values. Can just use a loop too rather than setting each type individually

tulir (approved):

PR #132, opened 2025-05-21
connector: support direct media

rnons:


adamvy:
this looks reasonable to me, merging for now so we can get direct media, but tulir can do an async review later

adamvy (approved):

PR #135, opened 2025-06-03
msgconv/from-meta: set mime type for instagram fetched media

1Conan:


tulir (approved):
If you've checked that the mime type is actually correct, then this looks fine

pkg/msgconv/from-meta.go
```diff
@@ -438,9 +442,11 @@ func (mc *MessageConverter) fetchFullXMA(ctx context.Context, att *table.Wrapped
 				minimalConverted.Extra["fi.mau.meta.xma_fetch_status"] = "reupload fail"
 				return minimalConverted
 			}
-			secondConverted.Content.Info.ThumbnailInfo = minimalConverted.Content.Info
-			secondConverted.Content.Info.ThumbnailURL = minimalConverted.Content.URL
-			secondConverted.Content.Info.ThumbnailFile = minimalConverted.Content.File
+			if !mc.DirectMedia {
+				secondConverted.Content.Info.ThumbnailInfo = minimalConverted.Content.Info
+				secondConverted.Content.Info.ThumbnailURL = minimalConverted.Content.URL
+				secondConverted.Content.Info.ThumbnailFile = minimalConverted.Content.File
```
tulir, line 448: There are 2 other blocks like this that need the if too

1Conan: fixed :D

tulir (approved):

PR #137, opened 2025-06-10
handlemeta: delete thread when we are removed as a participant

slice:


pkg/connector/handlemeta.go
```diff
@@ -354,11 +356,11 @@ func (m *MetaClient) handleDeleteThenInsertMessage(tk handlerParams, msg *table.
 	return wrapMessageDelete(tk.Portal, tk.UncertainReceiver, msg.MessageId)
 }
 
-func (m *MetaClient) handleDeleteThread(tk handlerParams, msg *table.LSDeleteThread) bridgev2.RemoteEvent {
+func (m *MetaClient) handleDeleteThreadKey(tk handlerParams, threadKey int64) bridgev2.RemoteEvent {
```
slice: This is a bit yucky and I initially wrote this to accept `ThreadKeyable` instead of `int64` but unfortunately that doesn't let you straightforwardly pass it to `handlePortalEvents` since the types don't work out… so I accept `int64` here and do a trivial trampoline when we get `LSDeleteThread` or `LSRemoveParticipantFromThread`, with the necessary condition attached to the latter.

pkg/connector/handlemeta.go
```diff
@@ -370,6 +372,10 @@ func (m *MetaClient) handleDeleteThread(tk handlerParams, msg *table.LSDeleteThr
 	}
```
tulir, line 371: Need to set `OnlyForMe: true` in the delete event since you leaving won't delete the group

slice: @tulir Updated to forward a `bool`, preserving the existing behavior of `OnlyForMe: false` for `LSDeleteThread`

tulir (approved):

PR #141, opened 2025-06-25
push: handle context cancellation and improve logging in ConnectBackground

bradtgmurray:
Got some logs with this ConnectBackground call not returning, adding more logs and adding better support for cancellation

### DIFF
--- a/pkg/connector/push.go
+++ b/pkg/connector/push.go
@@ -105,13 +105,18 @@ type connectBackgroundEvent struct {
 }
 
 func (m *MetaClient) ConnectBackground(ctx context.Context, params *bridgev2.ConnectBackgroundParams) error {
+	l := zerolog.Ctx(ctx)
+
 	evtChan := make(chan connectBackgroundEvent, 8)
 	m.connectBackgroundWAOfflineSync.Clear()
 	m.connectBackgroundEvt = evtChan
 	defer func() {
 		m.connectBackgroundEvt = nil
 	}()
+
 	go m.Connect(ctx)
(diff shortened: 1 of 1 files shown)


pkg/connector/push.go
```diff
@@ -105,13 +105,18 @@ type connectBackgroundEvent struct {
 }
 
 func (m *MetaClient) ConnectBackground(ctx context.Context, params *bridgev2.ConnectBackgroundParams) error {
+	l := zerolog.Ctx(ctx)
```
tulir, line 108: Usually the variable is called `log`

tulir (approved):

pkg/connector/push.go
bradtgmurray: I found I kept on colliding with the `log` package name so I've started calling it `l` 🤷

tulir: That's a feature: you should never import packages with the name `log`. If you see a collision, it means you have bad imports that have to be fixed

* The stdlib log isn't used anywhere and should never be imported
* The log subpackage in zerolog should only be imported in a very few cases. When it's really needed, it's better to alias it to something like `deflog` to make it clear that you mean to use the global logger. Almost all code should be using context loggers or loggers from structs, correct use of the global logger is very rare.

bradtgmurray: https://github.com/mautrix/meta/blob/d97005b8ab83228efe1f0a5e78af244d0ae6796f/pkg/messagix/events.go#L10 for example

Makes sense to me!

PR #143, opened 2025-07-01
Remove log in ConnectBackground that is way too verbose

bradtgmurray:


tulir:
Hmm, that shouldn't be way too verbose, I think the code is actually wrong

PR #146, opened 2025-07-17
login: ignore query parameters in wait for url pattern

1Conan:


pkg/connector/login.go
```diff
@@ -159,7 +159,7 @@ func (m *MetaCookieLogin) Start(ctx context.Context) (*bridgev2.LoginStep, error
 	case types.Messenger:
```
tulir, line 159: Should probably ignore on FB/IG too, not just messenger

tulir (approved):

PR #150, opened 2025-08-14
msgconv/from-whatsapp: Bridge replies to Facebook stories

radon-at-beeper:
BRI-32931

The previous behavior was that when a user replied to your Facebook story, the bridge would only display the URL of the story that was replied to, and would discard the actual reply message contents. With this change, the reply is shown instead of the URL, using a blockquote to provide context that this is a reply to a story. In the case that other types of messages can be submitted as replies to Facebook stories (e.g. images), it's unclear to me what would happen, but I think there's a chance things would just work.

The Messenger interface displays the message as an actual reply, but it's not a reply to a Messenger message, so this is not straightforward to replicate in the current data model.

### Facebook Messenger web interface

image

### Beeper desktop interface

image

### Bonus

I noticed that there are protobufs for responses to Instagram stories, too. Do we need to support those? I checked to see what happens, and no, those are already handled by the bridge as is:

image

### DIFF
--- a/pkg/msgconv/from-whatsapp.go
+++ b/pkg/msgconv/from-whatsapp.go
@@ -489,13 +489,13 @@ func (mc *MessageConverter) waExtendedContentMessageToMatrix(ctx context.Context
 			part.Content.EnsureHasHTML()
 			if nativeURL != "" {
 				part.Content.FormattedBody = fmt.Sprintf(
-					`<blockquote>Reply to <a href="%s">your Facebook story</a>:</blockquote><p>%s</p>`,
+					`<blockquote>Reply to <a href="%s">Facebook story</a>:</blockquote><p>%s</p>`,
 					nativeURL,
 					part.Content.FormattedBody,
 				)
 			} else {
 				part.Content.FormattedBody = fmt.Sprintf(
-					`<blockquote>Reply to your Facebook story:</blockquote><p>%s</p>`,
+					`<blockquote>Reply to Facebook story:</blockquote><p>%s</p>`,
 					part.Content.FormattedBody,
 				)
 			}


pkg/msgconv/from-whatsapp.go
```diff
@@ -459,6 +476,17 @@ func (mc *MessageConverter) waExtendedContentMessageToMatrix(ctx context.Context
...
 	}
+	switch content.GetTargetType() {
+	case waArmadilloXMA.ExtendedContentMessage_FB_STORY_REPLY:
+		parts, err := mc.waStoryReplyMessageToMatrix(ctx, content)
+		if err != nil {
+			break
+		}
+		for _, part := range parts {
+			part.Content.Body = "[Reply to your Facebook story] " + part.Content.Body
+		}
```
tulir, line 502: If we have the story URL, it'd be nice to keep that. Maybe something like `<blockquote>Reply to <a href="...">your Facebook story</a></blockquote>`

You can use `part.Content.EnsureHasHTML()` to make sure `FormattedBody` is filled, then prepend the html there

tulir: Oh and it should only add the text to one part (the part with type `MsgText` or `MsgNotice`, or if there is none, then add a new part)

radon-at-beeper: Thanks for the tips, new version looks like this:

image

pkg/msgconv/from-whatsapp.go
```diff
@@ -436,13 +438,30 @@ func (mc *MessageConverter) waLocationMessageToMatrix(ctx context.Context, conte
...
+	if err != nil {
+		return
+	}
+	parts = mc.waConsumerToMatrix(ctx, consMsg.GetPayload().GetContent())
+	return
+}
+
 func (mc *MessageConverter) waExtendedContentMessageToMatrix(ctx context.Context, content *waArmadilloXMA.ExtendedContentMessage) (parts []*bridgev2.ConvertedMessagePart) {
 	body := content.GetMessageText()
+	nativeUrl := ""
```
tulir: The Go naming convention includes consistent case for initialisms, so either `url` (if it's at the beginning) or `URL`, but never `Url` https://go.dev/wiki/CodeReviewComments#initialisms

radon-at-beeper: Oops. I was writing Kotlin lately and you're supposed to do it the cursed way over there, so I forgot the convention for Go.

tulir (approved):

PR #153, opened 2025-08-20
handle*: send and receive typing notifications for all Facebook chats

radon-at-beeper:
Implements BRI-33004.

E2EE Facebook chats:

https://github.com/user-attachments/assets/5f3d8472-7e23-40c7-8ba3-e6197b347b59

Non E2EE Facebook chats, which are totally different (lol):

https://github.com/user-attachments/assets/3f61fd09-ccb8-4a06-99c0-e49d67547f16

The implementation for E2EE chats is largely based on the WhatsApp bridge implementation, but adjusted for the Facebook environment. The biggest difference is the Facebook API doesn't use PushName, which requires https://github.com/tulir/whatsmeow/pull/906.

I also went ahead and dropped down the typing indicator timeout for non E2EE chats from 15 seconds to 5 seconds, the new value is more accurate to the official clients. The bridge was able to receive typing indicators from non E2EE chats, but wasn't sending its own. For E2EE chats, it wasn't doing either.

I tested community groups (threads), and apparently they don't support typing indicators at all in the official web client for Messenger, so I skipped over those.

### DIFF
--- a/pkg/connector/capabilities.go
+++ b/pkg/connector/capabilities.go
@@ -58,7 +58,7 @@ func supportedIfFFmpeg() event.CapabilitySupportLevel {
 }
 
 func capID() string {
-	base := "fi.mau.meta.capabilities.2025_04_21"
+	base := "fi.mau.meta.capabilities.2025_08_21"
 	if ffmpeg.Supported() {
 		return base + "+ffmpeg"
 	}


pkg/connector/handlematrix.go
```diff
@@ -516,3 +518,59 @@ func (m *MetaClient) HandleMatrixReadReceipt(ctx context.Context, receipt *bridg
...
+			zerolog.Ctx(ctx).Warn().Err(err).Msg("Failed to set presence when viewing chat")
+		}
+	}
+
+	// No codepaths where we return an error, yet
+	return nil
+}
+
+func (m *MetaClient) HandleMatrixTyping(ctx context.Context, msg *bridgev2.MatrixTyping) error {
+	portalJID := msg.Portal.Metadata.(*metaid.PortalMetadata).JID(msg.Portal.ID)
```
tulir: This looks like it needs to handle non-E2EE chats (which will probably give instagram support for free as well)

pkg/connector/handlematrix.go
```diff
@@ -516,3 +518,59 @@ func (m *MetaClient) HandleMatrixReadReceipt(ctx context.Context, receipt *bridg
...
+	case bridgev2.TypingTypeRecordingMedia:
+		mediaPresence = waTypes.ChatPresenceMediaAudio
+	case bridgev2.TypingTypeUploadingMedia:
+		return nil
+	}
+
+	err := m.updateWAPresence(waTypes.PresenceAvailable)
+	if err != nil {
+		zerolog.Ctx(ctx).Warn().Err(err).Msg("Failed to set presence on typing")
+	}
```
tulir: This should be behind a config option like it is in whatsapp

pkg/connector/capabilities.go
```diff
@@ -136,16 +136,17 @@ var metaCaps = &event.RoomFeatures{
...
+	Reply:               event.CapLevelFullySupported,
+	Edit:                event.CapLevelFullySupported,
+	EditMaxCount:        10,
+	EditMaxAge:          ptr.Ptr(jsontime.S(24 * time.Hour)),
+	Delete:              event.CapLevelFullySupported,
+	DeleteForMe:         false,
+	DeleteMaxAge:        ptr.Ptr(jsontime.S(10 * time.Minute)),
+	Reaction:            event.CapLevelFullySupported,
+	ReactionCount:       1,
+	TypingNotifications: true,
```
tulir, line 149: `capID` and `GetBridgeInfoVersion` need to be updated when capabilities are changed to tell the bridge it needs to resend the event to all rooms

pkg/connector/config.go
```diff
@@ -35,6 +35,9 @@ type Config struct {
 
 	DisplaynameTemplate string             `yaml:"displayname_template"`
 	displaynameTemplate *template.Template `yaml:"-"`
+
+	// Only affects E2EE chats right now.
+	SendPresenceOnTyping bool `yaml:"send_presence_on_typing"`
```
tulir, line 40: New config fields also need to be defined in `example-config.yaml` and `upgradeConfig` below

pkg/messagix/taskManager.go
```diff
@@ -41,6 +41,10 @@ func (tm *TaskManager) setTraceId(traceId string) {
 func (tm *TaskManager) AddNewTask(task socket.Task) {
 	payload, queueName, marshalQueueName := task.Create()
 	label := task.GetLabel()
+	if queueName == nil {
+		tm.client.Logger.Error().Any("label", label).Msg("no queue name provided for task")
```
tulir: `Any` should only be used for logging complicated or unknown types, if it's a known built-in type it's much more efficient to specify it

```suggestion
		tm.client.Logger.Error().Str("label", label).Msg("no queue name provided for task")
```

tulir (approved):

PR #154, opened 2025-08-22
handlematrix: Add nil check

radon-at-beeper:
Oops.

### DIFF
--- a/pkg/connector/handlematrix.go
+++ b/pkg/connector/handlematrix.go
@@ -524,12 +524,7 @@ func (m *MetaClient) HandleMatrixReadReceipt(ctx context.Context, receipt *bridg
 // whatsapp bridge, because they both use the same API.
 
 func (m *MetaClient) HandleMatrixViewingChat(ctx context.Context, msg *bridgev2.MatrixViewingChat) error {
-	if msg.Portal == nil {
-		return nil
-	}
-	portalMeta := msg.Portal.Metadata.(*metaid.PortalMetadata)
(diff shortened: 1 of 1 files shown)


pkg/connector/handlematrix.go
```diff
@@ -524,6 +524,9 @@ func (m *MetaClient) HandleMatrixReadReceipt(ctx context.Context, receipt *bridg
 // whatsapp bridge, because they both use the same API.
 
 func (m *MetaClient) HandleMatrixViewingChat(ctx context.Context, msg *bridgev2.MatrixViewingChat) error {
+	if msg.Portal == nil {
+		return nil
+	}
 	portalMeta := msg.Portal.Metadata.(*metaid.PortalMetadata)
 	if !portalMeta.ThreadType.IsWhatsApp() {
```
tulir: The comments below sound like this thread type check should be removed instead

radon-at-beeper: Actually the thread type check is still needed, the comment is outdated

radon-at-beeper: Removed the comment, we do only want to set online status for E2EE chat as it would have no effect for non E2EE chat. I was lazy when I wrote the original comment but later added the thread type check to actually check it properly.

tulir: Oh, well it should still go into the code below instead of returning immediately, with the current change it would never set presence to unavailable

radon-at-beeper: Oh, oops, I see what you mean. Refactored the whole thing to be correct.

PR #156, opened 2025-08-29
messagix: Receive Instagram typing indicators

radon-at-beeper:
Receive typing indicators from Instagram. This required reverse-engineering a totally novel websocket API that it seems nobody else on the Internet (that I was able to find) had published details of. Note that, surprisingly, typing indicators are sent via the same API on both Facebook and Instagram; however, they are only received using that same API on Facebook, instead being received via the other API on Instagram.

In `messagix/dgw` I have provided a reusable client for interacting with this API, which is able to encode and decode a subset of the available message frame types, enough to subscribe to typing indicators and parse those received. Received DGW messages are sent back to the bridge connector and handled in `handleMetaEvent` similarly to Lightspeed messages.

There are special ID schemas for Instagram, one used for user IDs and one used for thread IDs. Therefore, to interpret received typing indicators, the bridge must look up the corresponding FBIDs for these legacy IDs. The thread ID information was already being received as an additional message (`LSDeleteThenInsertIgThreadInfo`) that the bridge was not previously aware of, but the user ID information is only sometimes sent (in `LSDeleteThenInsertIGContactInfo`), and therefore I needed to update the bridge to react to `LSVerifyContactRowExists` by queuing a `GetContactsFullTask` for any contacts still missing information; this gives parity with the behavior of the web client. Instagram user and thread IDs are stored persistently in the attached database, since they are not expected to change over time. There are wrapper methods for handling both the in-memory and database caches of these ID types.

I wanted to separate out the DGW socket code into a separate module, but this required moving around a few other symbols that the Lightspeed socket code was accessing as private members. Most egregiously I am passing in a `MessagixClient` interface to the DGW module, since otherwise there would be an import cycle. Let me know if you can think of a way to improve the modularity.

The DGW socket code was originally based upon the Lightspeed socket code, but was rewritten to be more clearly correct to me from a concurrency perspective.

This has been tested to work with receiving Instagram typing notifications from another person who is typing in the Instagram web client, when I am using Beeper desktop with the self-hosted Instagram bridge running locally.

### DIFF
--- a/pkg/connector/handlemeta.go
+++ b/pkg/connector/handlemeta.go
@@ -172,7 +172,6 @@ func (m *MetaClient) handleMetaEvent(ctx context.Context, rawEvt any) {
 			if !evt.IsTyping {
 				timeout = 0
 			}
-			log.Error().Any("event", evt).Int64("thread_key", threadKey).Int64("user_id", userID).Msg("rrosborough: Activity indicator")
 			m.UserLogin.QueueRemoteEvent(&simplevent.Typing{
 				EventMeta: simplevent.EventMeta{
 					Type:      bridgev2.RemoteEventTyping,
@@ -245,13 +244,18 @@ func (m *MetaClient) handleParsedTable(ctx context.Context, isInitial bool, tbl 
 		}
 		m.syncGhost(ctx, contact)
 	}
+	for _, contact := range tbl.LSVerifyContactRowExists {
+		if ctx.Err() != nil {
+			return
+		}
+		m.syncGhost(ctx, contact)
+	}
 	if m.Client.Platform == types.Instagram {
 		contactsWithoutIGID := []int64{}
 		for _, contact := range tbl.LSVerifyContactRowExists {
 			if ctx.Err() != nil {
 				return
 			}
-			m.syncGhost(ctx, contact)
 			igid, err := m.getIGUserForFBID(ctx, contact.GetFBID())
 			if err != nil {
 				zerolog.Ctx(ctx).Warn().Err(err).Msg("Error getting IG user for FBID")
@@ -275,11 +279,13 @@ func (m *MetaClient) handleParsedTable(ctx context.Context, isInitial bool, tbl 
 				resp, err := m.Client.ExecuteTasks(ctx, tasks...)
 				if err != nil {
 					zerolog.Ctx(ctx).Warn().Err(err).Ints64("fbids", contactsBatch).Msg("user info request failed")
+					return
 				}
 				for _, info := range resp.LSDeleteThenInsertIGContactInfo {
 					err := m.putFBIDForIGUser(ctx, info.IgId, info.ContactId)
 					if err != nil {
 						zerolog.Ctx(ctx).Warn().Err(err).Msg("Failed to save FBID for IG user")
+						return
 					}
 				}
 				if len(contactsWithoutIGID) <= batchSize {


pkg/connector/handlemeta.go
```diff
@@ -203,11 +245,53 @@ func (m *MetaClient) handleParsedTable(ctx context.Context, isInitial bool, tbl
...
+		if err != nil {
+			zerolog.Ctx(ctx).Warn().Err(err).Msg("Error getting IG user for FBID")
+			continue
+		}
+		if igid == "" {
+			contactsWithoutIGID = append(contactsWithoutIGID, contact.GetFBID())
+		}
+	}
+	go func() {
+		for len(contactsWithoutIGID) > 0 {
```
tulir: All this looks like it should be limited to instagram clients and not be ran at all for messenger

pkg/metadb/00-latest-schema.sql
```diff
@@ -17,3 +17,35 @@ CREATE TABLE meta_reconnection_state (
...
+--
+-- Index on the fbid because occasionally we have to go the other way,
+-- for example looking up FB contacts on Instagram
+
+CREATE TABLE meta_instagram_user_id (
+  igid TEXT PRIMARY KEY,
+  fbid BIGINT NOT NULL
+);
+
+CREATE INDEX fbid_idx ON meta_instagram_user_id (fbid);
```
tulir: Should it be a unique key instead of just an index?

radon-at-beeper: Eh yeah, fair enough. Changed.

pkg/connector/handlemeta.go
radon-at-beeper: Oh yes good point. Changed.

pkg/connector/handlemeta.go
```diff
@@ -144,6 +147,45 @@ func (m *MetaClient) handleMetaEvent(ctx context.Context, rawEvt any) {
...
+			}
+			if userID == 0 {
+				log.Warn().Any("event", evt).Msg("Got activity indicator for unknown user ID")
+				return
+			}
+			timeout := 6 * time.Second
+			if !evt.IsTyping {
+				timeout = 0
+			}
+			log.Error().Any("event", evt).Int64("thread_key", threadKey).Int64("user_id", userID).Msg("rrosborough: Activity indicator")
```
tulir: Debug log

pkg/connector/handlemeta.go
```diff
@@ -203,11 +245,55 @@ func (m *MetaClient) handleParsedTable(ctx context.Context, isInitial bool, tbl
...
-	for _, contact := range tbl.LSVerifyContactRowExists {
-		if ctx.Err() != nil {
-			return
+	if m.Client.Platform == types.Instagram {
+		contactsWithoutIGID := []int64{}
+		for _, contact := range tbl.LSVerifyContactRowExists {
+			if ctx.Err() != nil {
+				return
+			}
+			m.syncGhost(ctx, contact)
```
tulir: This part still needs to run for messenger

pkg/connector/handlemeta.go
```diff
@@ -203,11 +245,55 @@ func (m *MetaClient) handleParsedTable(ctx context.Context, isInitial bool, tbl
...
+				contactsBatch := contactsWithoutIGID[:batchSize]
+				tasks := []socket.Task{}
+				for _, contact := range contactsBatch {
+					tasks = append(tasks, &socket.GetContactsFullTask{
+						ContactID: contact,
+					})
+				}
+				resp, err := m.Client.ExecuteTasks(ctx, tasks...)
+				if err != nil {
+					zerolog.Ctx(ctx).Warn().Err(err).Ints64("fbids", contactsBatch).Msg("user info request failed")
```
tulir, line 281: This should maybe return on error or something?

tulir (commented):
Otherwise looks good except slightly too much of the code moved inside `if platform == instagram`

pkg/connector/handlemeta.go
radon-at-beeper: fml thought I got all of those

pkg/connector/handlemeta.go
radon-at-beeper: Urgh yes

pkg/connector/handlemeta.go
radon-at-beeper: Fair enough, if the first one fails, the other ones would probably fail too

tulir (approved):

PR #157, opened 2025-09-02
Implement messenger-lite login method

radon-at-beeper:
Based originally on work done by @JJTech0130, but mostly rewritten since then.

If there's anything here that's not immediately clear, just let me know; it is easy and fast for me to add documentation and comments.

## New features

* New bridgev2 login flow, `messenger-lite`. For now, exposed only when the bridge is running in `meta` mode. Takes username (email) and password and handles the login noninteractively via the Messenger Lite API (Facebook Messenger iOS app). Meta calls the schema of this API "Bloks".
* Companion to `makeGraphQLRequest` called `makeBloksRequest`, which knows how to speak the idiosyncratic GraphQL dialect used by Bloks endpoints. Custom JSON marshaling supported for the structs used to supply input parameters, to transparently handle the numerous levels of json re-stringification that are needed by the Bloks API.
* Extensive custom JSON unmarshaling that sorts through the huge mess that is Bloks JSON and turns it into a nice tree structure of components, attributes, and scripts. There is a built-in pretty-printer for visualizing both code and document layout.
* Lisp parser that constructs an abstract syntax tree from embedded Bloks scripts. It's not very sophisticated but the edge cases have not proven important thus far.
* JSON file included in the source for undoing Bloks minification; covers components, attributes, and Lisp functions. The mapping is not complete, but partial unminification is supported and is sufficient to interpret large quantities of code. On a typical page, I would estimate that as much as 99.9% of the content is rendered correctly. The mapping was created through a combination of manual inspection and automated tooling.
* Lisp interpreter with large section of the Bloks standard library implemented, including facilities for control flow, data structure manipulation, variable and argument assignment, anonymous functions, parameter binding, RPC invocation, named procedure lookup, and handoff between multiple UI panes.
* Small command-line program `pkg/messagix/bloks/cmd` with a few optional flags that is good for invoking the Bloks pretty-printer and interpreter in various ways without having to pull in the whole bridge context. The code in there is not too important and is somewhat duplicative due to its main use being manual testing.
* Separate "bridge" struct which allows for configuring aspects of the Bloks runtime environment (e.g., providing return values for Bloks functions that query aspects of the mobile device, and accepting callbacks that ask to send an RPC or change which document is displayed).
* Selenium-like methods for locating descendant and ancestor components within a Bloks page, so that login can be implemented literally by filling text into the username and password fields, triggering their `on_text_changed` handlers, and then finding the `Log in` button and triggering its `touch` event handlers.
* Custom password encryption using the alternative cryptography needed by the Messenger Lite login flow, including fetching the public key used to encrypt the password to.

## Data flow changes

* Messenger Lite login requires a `DeviceID`. Allow generating one during the login process, then saving it on the client object for later installation into `WADevice.FacebookUUID` when E2EE is set up.
* Delegate to alternative logic in `EncryptPassword` when the platform is `MessengerLite`.
* New helper methods for Meta Cookies struct. The `UpdateValues` method now takes a map keyed by `MetaCookieName`.

## Refactoring

* New helper methods on the `PlatformType` enum that are now used instead of checking the enum types explicitly, so that adding new enum types does not automatically introduce bugs until you manually edit every callsite.

## Dependencies

* New dependency https://github.com/beeper/poly1305, a modified version of the Golang crypto standard library which allows using a nonstandard encryption algorithm ("legacy chacha20poly1305") needed for Messenger Lite login.

## New smells

* ~Some duplicated code in `MetaNativeLogin.SubmitUserInput` where it is the same as `MetaCookieLogin.SubmitCookies`. This should be coalesced. I will do so before PR is merged.~
* Some of the naming could be improved on the Bloks structs since there are so many of them all in the same package. There's also some unnecessary [Smurf naming](https://blog.codinghorror.com/new-programming-jargon/#21-smurf-naming-convention) with regard to `bloks.BloksXYZ`.
* Gross custom unmarshaler to parse `BloksResponseData` properly because the golang json library doesn't support keys with commas in them (yes, really). It's fixed in json/v2 so we can use that once it's ready.
* Error messages aren't great for issues around incorrect implementation of Bloks functions, due to the lack of stack traces and line numbering. This could be improved with some framework changes if it proves necessary.
* The parser and interpreter are both kind of gross and tend towards runtime errors if you make a coding mistake, because Go is exceptionally ill-suited to this kind of work at least out of the box. Typing could probably made more idiomatic to protect against more failure conditions.
* The Bloks tree unmarshaler is rather long-winded, and uses a somewhat fragile set of back-reference pointers to make the traversal methods work. Also, the `Unminify` method has been overloaded to do other kinds of important post-processing as well, which is unclean.

## Future work

* Two-factor authentication support (will be greatly eased by the Bloks interpreter here)
* Retrieve the APNS token after login

### DIFF
--- a/pkg/connector/login.go
+++ b/pkg/connector/login.go
@@ -14,12 +14,13 @@ import (
 	"maunium.net/go/mautrix/bridgev2/database"
 	"maunium.net/go/mautrix/bridgev2/status"
 
+	"go.mau.fi/util/exslices"
+
 	"go.mau.fi/mautrix-meta/pkg/messagix"
 	"go.mau.fi/mautrix-meta/pkg/messagix/cookies"
 	"go.mau.fi/mautrix-meta/pkg/messagix/types"
 	"go.mau.fi/mautrix-meta/pkg/messagix/useragent"
 	"go.mau.fi/mautrix-meta/pkg/metaid"
-	"go.mau.fi/util/exslices"
 )
 
 const (


Fizzadar (approved):
This is absolutely wild 😆 Looks good from a quick scan (I've not reviewed the bloks module in detail), excited for 2FA + push reg.

pkg/connector/handlematrix.go
```diff
@@ -692,7 +692,7 @@ func (m *MetaClient) HandleMatrixRoomName(ctx context.Context, msg *bridgev2.Mat
 			return false, err
 		}
 		return true, nil
-	} else if platform == types.Facebook || platform == types.Messenger {
+	} else if platform == types.Facebook || platform == types.Messenger || platform == types.MessengerLite {
```
tulir: This is probably supposed to apply for all messenger types (the old code was likely wrong for excluding facebook-tor)

```suggestion
	} else {
```

pkg/connector/login.go
```diff
@@ -269,3 +284,115 @@ func (m *MetaCookieLogin) SubmitCookies(ctx context.Context, strCookies map[stri
...
+		},
+		Metadata: &metaid.UserLoginMetadata{
+			Platform: client.GetCookies().Platform,
+			Cookies:  client.GetCookies(),
+			LoginUA:  loginUA,
+		},
+	}, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to save new login: %w", err)
+	}
```
tulir: This part (and maybe some after too) should probably be deduplicated with cookie login

pkg/connector/handlematrix.go
radon-at-beeper: Noted. Updating to use `platform.IsMessenger()` instead, should be equivalent and more clear.

pkg/connector/login.go
radon-at-beeper: Agree, I mentioned this in the PR description, am coalescing before merge.

pkg/messagix/bloks/interp.go
```diff
@@ -0,0 +1,765 @@
...
+	call, ok := form.BloksScriptNodeContent.(*BloksScriptFuncall)
+	if !ok {
+		return nil, fmt.Errorf("unexpected script node %T", form.BloksScriptNodeContent)
+	}
+	// Some of the cases in this switch are not needed for any given login. However different
+	// functions get pulled in depending on which API you are talking to, so I left in
+	// everything that came up at one point or another during testing.
+	switch call.Function {
+	case "bk.action.core.If":
+		{
```
tulir, line 175: Why are all these nested in an extra block 🤔

radon-at-beeper: Cuz in golang all the switch cases share a scope and therefore you cannot reuse the same variable name in two different cases even if there is no fallthrough :roll_eyes:

radon-at-beeper: Correction (by tulir) this is completely not true I don't know why I thought this. Fixing in follow-up PR

PR #158, opened 2025-09-09
msgconv: Fix multi-image direct media

radon-at-beeper:
When using direct media mode (for Beeper, this means a local bridge as opposed to cloud or self-hosted), messages with N image attachments would instead render as N copies of the first image attachment, when received in a group chat (as opposed to a DM).

This is because the image attachments were transformed into direct media objects, which are given unique IDs that encode information used to look up metadata for them later. However, the unique ID was only based on the message ID, leading to all attachments from the same message being treated identically. In the code path where message IDs were decoded and actioned upon, there was a suspicious use of `GetFirstPartByID` rather than `GetPartByID`, which resulted in only the first attachment being used in each case.

New enum values for direct media types are introduced, which support encoding a part ID alongside the message ID, and the old ones are deprecated (but support is retained). The code for parsing media IDs was refactored to be easier to understand, due to the addition of more code paths. Finally, calling code was adjusted to attach appropriate unique values for message part IDs to the context, so that it can be used when constructing media IDs if present. This context augmentation only needs to happen in cases where there may be multiple media attachments in the same message. Since that does not appear to be possible in WhatsApp (multiple images can be sent at once, but are processed as separate messages), only the Meta code paths are modified.

This solves BRI-33018 and related tickets.

### DIFF
--- a/pkg/msgconv/from-meta.go
+++ b/pkg/msgconv/from-meta.go
@@ -96,14 +96,17 @@ func (mc *MessageConverter) ToMatrix(
 	if msg.IsUnsent {
 		return cm
 	}
-	for _, blobAtt := range msg.BlobAttachments {
+	for i, blobAtt := range msg.BlobAttachments {
+		ctx := context.WithValue(ctx, contextKeyPartID, networkid.PartID(fmt.Sprintf("blob_attachment_%d", i)))
 		cm.Parts = append(cm.Parts, mc.blobAttachmentToMatrix(ctx, blobAtt))
 	}
-	for _, legacyAtt := range msg.Attachments {
+	for i, legacyAtt := range msg.Attachments {
+		ctx := context.WithValue(ctx, contextKeyPartID, networkid.PartID(fmt.Sprintf("attachment_%d", i)))
 		cm.Parts = append(cm.Parts, mc.legacyAttachmentToMatrix(ctx, legacyAtt))
 	}
 	var urlPreviews []*table.WrappedXMA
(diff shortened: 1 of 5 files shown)


pkg/msgconv/from-meta.go
```diff
@@ -96,14 +96,17 @@ func (mc *MessageConverter) ToMatrix(
 	if msg.IsUnsent {
 		return cm
 	}
-	for _, blobAtt := range msg.BlobAttachments {
+	for i, blobAtt := range msg.BlobAttachments {
+		ctx := context.WithValue(ctx, contextKeyPartID, networkid.PartID(fmt.Sprintf("blob_attachment_%d", i)))
```
tulir, line 100: Was there some particular reason for changing these? (it's probably fine, just wondering)

radon-at-beeper: ~Oh... you know what... I initially thought part IDs weren't being assigned to the messages and so I would need to add them later. But then I discovered there was a separate codepath further down where they were being assigned, and it was just the lookup code wasn't using them. So I can and should get rid of this. Thanks~

radon-at-beeper: Oh wait no I remember. The issue is the part IDs are assigned later, after the message fragments are already created. But with the way the media ID generation works, the message part IDs need to be embedded when the IDs are generated, which has already happened. So if we want part IDs to get embedded properly, we need to assign the part IDs ahead of time. So this is needed.

PR #165, opened 2025-10-23
msgconv/from-whatsapp: add better error message for unsupported stickers

radon-at-beeper:
When receiving a sticker message that uses receiver fetch, which isn't yet supported (requires extensive novel cryptography), display "Unsupported sticker, view in Messenger" rather than "Failed to transfer media". Also incorporate https://github.com/tulir/whatsmeow/pull/978 to display the error right away instead of after a long delay of network retries. I confirmed that supported stickers (e.g. sent through the web interface) do not hit this case, and are still displayed correctly.

### DIFF
--- a/pkg/msgconv/from-whatsapp.go
+++ b/pkg/msgconv/from-whatsapp.go
@@ -26,8 +26,6 @@ import (
 	"net/url"
 	"strings"
 	"time"
-	"unicode"
-	"unicode/utf8"
 
 	"github.com/rs/zerolog"
 	"go.mau.fi/util/exmime"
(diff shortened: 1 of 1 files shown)


pkg/msgconv/from-whatsapp.go
```diff
@@ -120,6 +122,10 @@ type AttachmentMessageWithCaption[Integral MediaTransportContainer, Ancillary an
 
 type convertFunc func(ctx context.Context, data []byte, mimeType string) ([]byte, string, string, error)
 
+type userVisibleError struct {
+	error
+}
```
tulir, line 129: It'd be better to just have an `Message string` inside the struct and implement the error interface yourself. That way you also avoid the linter error complaining about uppercase error messages, so you won't need the hacky capitalization afterwards

radon-at-beeper: Oh, well I was assuming that we had the linter enabled because we actually did want all our error objects to be lowercase. You think it'd be better to just bypass the linter in this case though?

tulir: It's a Go convention because errors are often nested (as in `something bad happened: <nested error message>`), but that never happens for this particular error

radon-at-beeper: OK, cool, I changed it

tulir (approved):

PR #167, opened 2025-10-30
msgconv/from-matrix: Fallback to mobile API for video uploads

radon-at-beeper:
As described in the code comments. The new request is only made if all of the following conditions apply:
1. The bridge is connected to Instagram (not Facebook), and
2. The upload is a video (not an image, audio, or file), and
3. The regular request through the web API received a successful response, but
4. The successful response did not contain a valid attachment ID

... indicating the failure condition which prevents some Instagram accounts from uploading video attachments through the Instagram web API even when using the official Instagram website.

If the fallback is successful, an INFO log is printed. If the fallback also fails, an additional WARN log is printed. Note that the latency of the fallback upload API is quite high, even when using the official Instagram app.

### DIFF
--- a/pkg/msgconv/from-matrix.go
+++ b/pkg/msgconv/from-matrix.go
@@ -280,7 +280,7 @@ func (mc *MessageConverter) reuploadFileToMeta(ctx context.Context, client *mess
 		}
 	}
 	if attachmentID == 0 {
-		return 0, fmt.Errorf("failed to upload attachment: fbid not received")
+		return 0, fmt.Errorf("%w: fbid not received", bridgev2.ErrMediaReuploadFailed)
 	}
 	return attachmentID, nil
 }


pkg/msgconv/from-matrix.go
```diff
@@ -269,3 +287,86 @@ func (mc *MessageConverter) reuploadFileToMeta(ctx context.Context, client *mess
...
+	h.Add("x-fb-client-ip", "True")
+	h.Add("x-fb-friendly-name", "undefined:media-upload")
+	h.Add("x-fb-http-engine", "Tigon/MNS/TCP")
+	h.Add("x-fb-rmd", "state=URL_ELIGIBLE")
+	h.Add("x-fb-server-cluster", "True")
+	h.Add("x-zero-balance", "INIT")
+	h.Add("x-zero-eh", "")
+	resp, body, err := client.MakeRequest(
+		ctx,
+		fmt.Sprintf("https://rupload.facebook.com/messenger_video/%s", uploadID),
```
tulir, line 348: I think `messenger_video` will only work for videos, should probably either only use this for `m.video` messages or make the url dynamic

pkg/msgconv/from-matrix.go
```diff
@@ -269,3 +287,86 @@ func (mc *MessageConverter) reuploadFileToMeta(ctx context.Context, client *mess
...
+}
+
+type ruploadResponse struct {
+	ID      int   `json:"id"`
+	MediaID int64 `json:"media_id"`
+}
+
+func (mc *MessageConverter) reuploadFileToMetaFallback(ctx context.Context, client *messagix.Client, content *event.MessageEventContent) (int64, error) {
+	mime := content.Info.MimeType
+	data, err := mc.Bridge.Bot.DownloadMedia(ctx, content.URL, content.File)
```
tulir: It would be preferable not to download the file twice

pkg/msgconv/from-matrix.go
radon-at-beeper: Yes, I have configured it in the calling function to only use the fallback in case of video messages, it is skipped for any other message type.

tulir: Oops yeah I missed that

pkg/msgconv/from-matrix.go
radon-at-beeper: Refactored to download the file once.

pkg/msgconv/from-matrix.go
radon-at-beeper: Renamed the method to make it clear that it can only be used for videos.

pkg/msgconv/from-matrix.go
```diff
@@ -265,7 +264,103 @@ func (mc *MessageConverter) reuploadFileToMeta(ctx context.Context, client *mess
...
+	}
+	attachmentID := resp.Payload.RealMetadata.GetFbId()
+	if attachmentID == 0 {
+		zerolog.Ctx(ctx).Warn().RawJSON("response", resp.Raw).Msg("No fbid received for upload")
+	}
+	if attachmentID == 0 && content.MsgType == event.MsgVideo && client.Platform == types.Instagram {
+		attachmentID, err = mc.reuploadVideoToMetaFallback(ctx, client, data, mime)
+		if err != nil {
+			zerolog.Ctx(ctx).Warn().Err(err).Msg("Failed to upload attachment via Instagram Android fallback")
+			return 0, err
```
tulir:
```suggestion
			return 0, fmt.Errorf("%w: fallback upload failed: %w", bridgev2.ErrMediaReuploadFailed, err)
```

pkg/msgconv/from-matrix.go
```diff
@@ -265,7 +264,103 @@ func (mc *MessageConverter) reuploadFileToMeta(ctx context.Context, client *mess
...
+		} else {
+			zerolog.Ctx(ctx).Info().Msg("Uploaded attachment via Instagram Android fallback")
+		}
+	}
+	if attachmentID == 0 {
+		return 0, fmt.Errorf("failed to upload attachment: fbid not received")
```
tulir: This should actually also use the wrapped error

```suggestion
		return 0, fmt.Errorf("%w: fbid not received", bridgev2.ErrMediaReuploadFailed)
```

tulir (approved):

PR #176, opened 2025-12-04
client: always pick worst-case bridge state

Fizzadar:
This avoids state flapping between two values by consistently picking the worst case state.

### DIFF
--- a/pkg/connector/client.go
+++ b/pkg/connector/client.go
@@ -555,21 +555,16 @@ func (m *MetaClient) resetWADevice() {
 }
 
 func (m *MetaClient) FillBridgeState(state status.BridgeState) status.BridgeState {
-	if state.StateEvent == status.StateConnected || state.Error == WADisconnected {
-		var copyFrom *status.BridgeState
-		if m.waState.StateEvent != "" && m.waState.StateEvent != status.StateConnected {
-			copyFrom = &m.waState
-		}
-		if m.metaState.StateEvent != "" && m.metaState.StateEvent != status.StateConnected {
-			copyFrom = &m.metaState
-		}
-		if copyFrom != nil {
-			state.StateEvent = copyFrom.StateEvent
-			state.Error = copyFrom.Error
-			state.Message = copyFrom.Message
-			state.Info = copyFrom.Info
-		}
+	// The Meta bridge internally has two states - one for connection to meta and one for whatsapp;
+	// as such we need to merge the two states when sending updates and pick the "worst case".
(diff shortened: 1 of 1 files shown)


pkg/connector/client.go
```diff
@@ -580,6 +569,31 @@ func (m *MetaClient) FillBridgeState(state status.BridgeState) status.BridgeStat
...
+	if state.StateEvent == m.waState.StateEvent && state.StateEvent == m.metaState.StateEvent {
+		// If both states are the same as the input, we can use as-is
+		return state
+	}
+
+	// Now find the worst case state in order (UNKNOWN_ERROR being worst), prefer the input state
+	// if matches or fallback to either metaState or waState.
+	for _, status := range []status.BridgeStateEvent{
+		status.StateUnknownError,
+		status.StateBadCredentials,
```
tulir, line 587: Bad credentials should probably be worse 🤔

also don't forget connecting

Fizzadar: Yeah I think that's right, updated

pkg/connector/client.go
```diff
@@ -580,6 +569,32 @@ func (m *MetaClient) FillBridgeState(state status.BridgeState) status.BridgeStat
 	return state
 }
 
+func (m *MetaClient) pickWorstCaseBridgeState(state status.BridgeState) status.BridgeState {
+	if state.StateEvent == m.waState.StateEvent && state.StateEvent == m.metaState.StateEvent {
+		// If both states are the same as the input, we can use as-is
+		return state
+	}
+
+	// Now find the worst case state in order (UNKNOWN_ERROR being worst), prefer the input state
```
tulir: Now the comment is outdated

pkg/connector/client.go
```diff
@@ -555,21 +555,10 @@ func (m *MetaClient) resetWADevice() {
...
-		}
-		if m.metaState.StateEvent != "" && m.metaState.StateEvent != status.StateConnected {
-			copyFrom = &m.metaState
-		}
-		if copyFrom != nil {
-			state.StateEvent = copyFrom.StateEvent
-			state.Error = copyFrom.Error
-			state.Message = copyFrom.Message
-			state.Info = copyFrom.Info
-		}
```
tulir, line 574: This part was intentional to avoid overwriting other fields like remote profiles and such

Fizzadar: Ah good flag, re-done it with the same copy technique so we don't mangle those.

PR #178, opened 2025-12-05
msgconv: Fix sticker sizes and thumbs-up rendering

radon-at-beeper:
PLAT-31683, three issues here:
1. ~For some reason we were ignoring `PreviewWidth` and `PreviewHeight` for animated media. We weren't using different attributes for those media, we were just setting the width and height to zero. I don't understand why the original code was written the way it was, any context appreciated. I've updated it to use these attributes when available.~
2. It appears that Facebook actually renders all stickers at 96x96 pixels even though this doesn't match any of the width or height attributes associated with the data that comes in from the server. I figure we should just hardcode to match. I've added a code comment with more details. I wonder if this would have any adverse effects (e.g. maybe the logic is actually more complicated), feedback on suggested testing appreciated.
3. We were bridging the thumbs-up button as an emoji in E2EE chats but a sticker in non-E2EE chats. I've changed it to an emoji for all chats as that's more similar to what Facebook does (they don't actually render a sticker, and it's rendered at a different size than a sticker too). This change I feel pretty good about, but let me know if the implementation is bad.

## Messenger web client:

image

## Bridged - after this change:

image

## Bridged - before this change:

image

### DIFF
--- a/pkg/msgconv/from-meta.go
+++ b/pkg/msgconv/from-meta.go
@@ -324,12 +324,14 @@ func (mc *MessageConverter) legacyAttachmentToMatrix(ctx context.Context, att *t
 		mime = att.AttachmentMimeType
 	}
 	duration := att.PlayableDurationMs
+	var width, height int64
 	if url == "" {
 		url = att.PreviewUrl
 		mime = att.PreviewUrlMimeType
+		width, height = att.PreviewWidth, att.PreviewHeight
 	}
 	converted, err := mc.reuploadAttachment(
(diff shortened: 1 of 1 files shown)


radon-at-beeper:
Should zeroing out the dimensions on the bridge level instruct clients to read the dimensions from the actual media file? This doesn't seem to be what our current clients are doing, as they display stickers scaled-up and noticeably pixelated.

radon-at-beeper:
Got it, I will file a separate bug for that then, and I have reverted change 1. Let me know what you think of the remainder.

PR #188, opened 2026-01-07
directmedia: implement refreshing URLs

Fizzadar:
Used claude to do this but based on prior discussions the approach seems correct. Attaching the investigation notes + plan generated. Basically: expand the direct media object to include enough information to enable refreshing the URL if the current one has expired.

[investigation-notes.md](https://github.com/user-attachments/files/24470317/investigation-notes.md)
[investigation-plan.md](https://github.com/user-attachments/files/24470318/investigation-plan.md)

### DIFF
--- a/pkg/msgconv/from-meta.go
+++ b/pkg/msgconv/from-meta.go
@@ -486,8 +486,8 @@ func (mc *MessageConverter) xmaLocationToMatrix(ctx context.Context, att *table.
 	}
 }
 
-var ReelActionURLRegex = regexp.MustCompile(`^/stories/direct/(\d+)_(\d+)$`)
-var ReelActionURLRegex2 = regexp.MustCompile(`^https://instagram\.com/stories/([a-z0-9.-_]{3,32})/(\d+)$`)
+var reelActionURLRegex = regexp.MustCompile(`^/stories/direct/(\d+)_(\d+)$`)
+var reelActionURLRegex2 = regexp.MustCompile(`^https://instagram\.com/stories/([a-z0-9.-_]{3,32})/(\d+)$`)
 var usernameRegex = regexp.MustCompile(`^[a-z0-9.-_]{3,32}$`)
 var ErrURLNotFound = errors.New("url not found")
 
@@ -586,7 +586,7 @@ func (mc *MessageConverter) fetchFullXMA(ctx context.Context, att *table.Wrapped
 	case strings.HasPrefix(att.CTA.ActionUrl, "/stories/direct/"):
 		log.Trace().Any("cta_data", att.CTA).Msg("Fetching XMA story from CTA data")
 		externalURL := fmt.Sprintf("https://www.instagram.com%s", att.CTA.ActionUrl)
-		match := ReelActionURLRegex.FindStringSubmatch(att.CTA.ActionUrl)
+		match := reelActionURLRegex.FindStringSubmatch(att.CTA.ActionUrl)
 		if usernameRegex.MatchString(att.HeaderTitle) && len(match) == 3 {
 			// Very hacky way to hopefully fix the URL to work on mobile.
 			// When fetching the XMA data, this is done again later in a safer way.
@@ -691,7 +691,7 @@ func (mc *MessageConverter) fetchFullXMA(ctx context.Context, att *table.Wrapped
 			return minimalConverted
 		}
 
-		if match := ReelActionURLRegex2.FindStringSubmatch(att.CTA.ActionUrl); len(match) != 3 {
+		if match := reelActionURLRegex2.FindStringSubmatch(att.CTA.ActionUrl); len(match) != 3 {
 			log.Warn().Str("action_url", att.CTA.ActionUrl).Msg("Failed to parse story action URL (type 2)")
 			minimalConverted.Extra["fi.mau.meta.xma_fetch_status"] = "parse fail"
 			return minimalConverted


pkg/connector/directmedia.go
```diff
@@ -85,3 +117,238 @@ func (m *MetaConnector) Download(ctx context.Context, mediaID networkid.MediaID,
...
+
+	// Collect all messages to search through
+	var allMessages []*table.WrappedMessage
+	for _, upsert := range upsertMap {
+		allMessages = append(allMessages, upsert.Messages...)
+	}
+	allMessages = append(allMessages, insertMsgs...)
+
+	for _, wrappedMsg := range allMessages {
+		if wrappedMsg.MessageId != fbMsgID.ID {
```
Fizzadar: This is wasteful. We should instead modify the `DBMetadata` for each message with updated URLs as we go and then return the one we're interested in.

Fizzadar: https://github.com/mautrix/meta/pull/188/commits/13fbd989b062f212f77b754085664e8c82ea22e8

pkg/msgconv/from-meta.go
```diff
@@ -818,10 +885,22 @@ func (mc *MessageConverter) xmaAttachmentToMatrix(ctx context.Context, att *tabl
...
 }
 
+// MediaRefreshMeta contains identifiers needed to refresh expired media URLs
+type MediaRefreshMeta struct {
+	ExpiresAt      int64  // Unix ms timestamp when URL expires
+	AttachmentFbid string // For blob attachments
+	PartIndex      int    // For blob attachments (fallback matching)
+	XMATargetId    int64  // For XMA attachments (Instagram API)
+	XMAShortcode   string // For XMA attachments (Instagram API)
+	XMAActionUrl   string // For XMA story attachments
```
tulir: Id/Url -> ID/URL

pkg/msgconv/media.go
```diff
@@ -203,8 +203,19 @@ func downloadChunkedVideo(ctx context.Context, mime, url string, maxSize int64)
...
+	ExpiresAt int64  `json:"expires_at,omitempty"` // Unix ms timestamp
+
+	// For blob attachments (message re-fetch):
+	AttachmentFbid string `json:"attachment_fbid,omitempty"`
+	PartIndex      int    `json:"part_index,omitempty"`
+
+	// For XMA attachments (Instagram API refresh):
+	XMATargetId  int64  `json:"xma_target_id,omitempty"`
+	XMAShortcode string `json:"xma_shortcode,omitempty"`
+	XMAActionUrl string `json:"xma_action_url,omitempty"` // for stories
```
tulir: Also casing here

pkg/connector/directmedia.go
```diff
@@ -54,7 +60,33 @@ func (m *MetaConnector) Download(ctx context.Context, mediaID networkid.MediaID,
...
+		url := info.URL
+
+		// Check if URL is expired or about to expire (5 minute buffer)
+		needsRefresh := info.ExpiresAt > 0 && time.Now().UnixMilli() > info.ExpiresAt-5*60*1000
+
+		// Try download first
+		size, reader, err := msgconv.DownloadMedia(ctx, info.MimeType, url, m.MsgConv.MaxFileSize)
+
+		// If download failed with 403 or we know URL is expired, try to refresh
+		if err != nil && (strings.Contains(err.Error(), "403") || needsRefresh) {
```
tulir: Error string comparisons aren't very nice

pkg/connector/directmedia.go
```diff
@@ -54,7 +60,33 @@ func (m *MetaConnector) Download(ctx context.Context, mediaID networkid.MediaID,
...
+			log.Debug().
+				Bool("needs_refresh", needsRefresh).
+				Str("original_error", err.Error()).
```
tulir:
```suggestion
				Err("original_error", err).
```

pkg/connector/directmedia.go
```diff
@@ -85,3 +117,312 @@ func (m *MetaConnector) Download(ctx context.Context, mediaID networkid.MediaID,
...
+	if ig == nil {
+		return "", fmt.Errorf("instagram client not available")
+	}
+
+	log := zerolog.Ctx(ctx)
+
+	// Handle story refresh (type 1: /stories/direct/...)
+	if info.XMAActionUrl != "" && strings.HasPrefix(info.XMAActionUrl, "/stories/direct/") {
+		match := msgconv.ReelActionURLRegex.FindStringSubmatch(info.XMAActionUrl)
+		if len(match) == 3 {
```
tulir: It could probably just store the parsed data rather than re-parsing on refresh 🤔

pkg/connector/directmedia.go
```diff
@@ -85,3 +117,312 @@ func (m *MetaConnector) Download(ctx context.Context, mediaID networkid.MediaID,
...
+	return "", fmt.Errorf("target attachment not found in re-fetched messages")
+}
+
+// extractBestURL selects the highest resolution URL from Instagram media response
+func extractBestURL(item *responses.Items, mediaType string) (string, error) {
+	var bestURL string
+	var bestRes int
+
+	// Prefer video versions if media type is video or if video versions exist
+	if mediaType == "video" || mediaType == "story" || len(item.VideoVersions) > 0 {
```
tulir: This method is a bit sus. Is mediaType really needed at all?

Fizzadar: Yeah I think we can just always use the highest res video and then fallback to image.

Fizzadar: Which also allows us to drop the field entirely and a bunch of unnecessary logic :)

pkg/connector/directmedia.go
Fizzadar: Yep way nicer

tulir (approved):

PR #193, opened 2026-01-09
client: fallback to non-cached login if we fail to authenticate

Fizzadar:
Should address `messagix-socket: client has not been authenticated successfully yet` errors when land the bridge stuck.


tulir (approved):
Sounds pretty sus that a cached login wouldn't have the necessary data, but maybe this is fine

PR #194, opened 2026-01-12
connector: re-request history backfill before timeout

Fizzadar:
This seems to be a bug in the implementation where sync on reconnect indicates more messages are available on the server between the most recent we have and the insertion batch, we need to fetch those to fill the gap. Based on the logs it looks like for some threads we get 10 max messages and others 20.

### DIFF
--- a/pkg/connector/backfill.go
+++ b/pkg/connector/backfill.go
@@ -243,7 +243,6 @@ func (m *MetaClient) FetchMessages(ctx context.Context, params bridgev2.FetchMes
 			return nil, fmt.Errorf("backfill collector already exists for thread %d", threadID)
 		}
 		defer m.removeBackfillCollector(threadID, collector)
-		// Figure out timeout, ticker 1/10th of that
 		start := time.Now()
 		timeout := BackfillTimeout
 		if params.Forward && bridgev2.PortalEventBuffer == 0 {


pkg/connector/backfill.go
```diff
@@ -219,6 +225,10 @@ func (m *MetaClient) FetchMessages(ctx context.Context, params bridgev2.FetchMes
 			zerolog.Ctx(ctx).Warn().Msg("Can't backfill chat with no messages")
 			return nil, nil
 		}
+		if params.Forward && upsert.Range.MinTimestampMs >= oldestMessageTS {
+			zerolog.Ctx(ctx).Warn().Msg("Forward backfill has all messages already")
+			return nil, nil
```
tulir: This does need to actually return the data like it does at the end of the function. Also I think we don't want this check for the fake upsert data, only for the bundled one with real messages included

```suggestion
		if params.Forward && params.BundledData != nil && upsert.Range.MinTimestampMs >= oldestMessageTS {
			zerolog.Ctx(ctx).Debug().Msg("Forward backfill has all messages already")
			return m.wrapBackfillEvents(ctx, params.Portal, upsert, params.AnchorMessage, params.Forward), nil
```

pkg/connector/backfill.go
```diff
@@ -232,52 +242,47 @@ func (m *MetaClient) FetchMessages(ctx context.Context, params bridgev2.FetchMes
...
-					newMinTS = collector.UpsertMessages.Range.MinTimestampMs
-				}
-				if newMinTS == 0 || newMinTS == prevMinTS {
-					zerolog.Ctx(ctx).Error().
-						Any("received_range", collector.UpsertMessages.Range).
-						Msg("Waiting for backfill collector timed out")
-					m.removeBackfillCollector(threadID, collector)
-					upsert = collector.UpsertMessages
+				if params.Forward && collector.UpsertMessages.Range.MinTimestampMs >= oldestMessageTS {
+					zerolog.Ctx(ctx).Warn().Msg("Backfill collector did not call done, but has all messages")
```
tulir, line 268: This check might also need to exclude the fake range when there's no bundled data

Fizzadar: Looks like bundled data is required when doing forward backfill https://github.com/mautrix/meta/blob/main/pkg/connector/backfill.go#L186

tulir: Oh right, then it's fine as is

pkg/connector/backfill.go
```diff
@@ -232,52 +242,47 @@ func (m *MetaClient) FetchMessages(ctx context.Context, params bridgev2.FetchMes
 		if !m.addBackfillCollector(threadID, collector) {
 			return nil, fmt.Errorf("backfill collector already exists for thread %d", threadID)
 		}
-		if !m.requestMoreHistory(ctx, threadID, oldestMessageTS, oldestMessageID) {
-			m.removeBackfillCollector(threadID, collector)
-			return nil, fmt.Errorf("failed to request more history for thread %d", threadID)
-		}
-		timeout := 30 * time.Second
+		defer m.removeBackfillCollector(threadID, collector)
+		// Figure out timeout, ticker 1/10th of that
```
tulir:
```suggestion
```

PR #196, opened 2026-01-13
connector,messagix: support backfill of all chats

Fizzadar:
Chats, not messages.

### DIFF
--- a/pkg/connector/threadbackfill.go
+++ b/pkg/connector/threadbackfill.go
@@ -31,7 +31,7 @@ func (m *MetaClient) runThreadBackfill(ctx context.Context) error {
 	delay := m.Main.Config.ThreadBackfill.BatchDelay
 
 	batchCount := 0
-	lastMinThreadKey := m.Client.GetSyncGroupKeyStore(1).MinThreadKey
+	var lastMinThreadKey int64
 
 	for {
 		if ctx.Err() != nil {


pkg/connector/threadbackfill.go
```diff
@@ -0,0 +1,95 @@
...
+
+	batchCount := 0
+	lastMinThreadKey := m.Client.GetSyncGroupKeyStore(1).MinThreadKey
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
```
tulir: This is equivalent when there's no waiting being done:

```suggestion
		if ctx.Err() != nil {
			return ctx.Err()
		}
```

pkg/connector/threadbackfill.go
```diff
@@ -0,0 +1,95 @@
...
+
+	return m.runThreadBackfill(ctx)
+}
+
+func (m *MetaClient) runThreadBackfill(ctx context.Context) error {
+	log := zerolog.Ctx(ctx)
+	delay := m.Main.Config.ThreadBackfill.BatchDelay
+
+	batchCount := 0
+	lastMinThreadKey := m.Client.GetSyncGroupKeyStore(1).MinThreadKey
```
tulir: Did resuming pagination not work or did we just decide to restart it every time?

pkg/connector/example-config.yaml
```diff
@@ -50,3 +50,10 @@ send_presence_on_typing: false
 receive_instagram_typing_indicators: true
 # Should view-once messages be disabled entirely?
 disable_view_once: false
+
+# Thread backfill settings for syncing older conversations
+thread_backfill:
+  # Enable thread list backfill to sync older conversations on connect
+  enabled: false
```
tulir: Batch count limit would probably be more appropriate than an enabled boolean (-1 for unlimited, 0 for disabled, >0 for limited)

pkg/connector/threadbackfill.go
```diff
@@ -0,0 +1,95 @@
...
+
+		if tbl == nil {
+			log.Info().Int("batches_processed", batchCount).Msg("Thread backfill complete - no more threads")
+			m.markBackfillComplete(ctx, m.LoginMeta)
+			return nil
+		}
+
+		batchCount++
+
+		keyStore := m.Client.GetSyncGroupKeyStore(1)
```
tulir: Fetching this inside FetchMoreThreads and again here is a bit weird

pkg/connector/threadbackfill.go
Fizzadar: Didn't seem to work because the order isn't consistent (!), so unfortunately need to paginate it all to be sure.

tulir: 🙀

pkg/connector/threadbackfill.go
Fizzadar: 🤦 there used to be another thing here!

pkg/connector/threadbackfill.go
Fizzadar: Yeah should just come back from fetch more => https://github.com/mautrix/meta/pull/196/commits/4281417894ecbda8e2721f7f02a96ab418d3dd4c

pkg/connector/example-config.yaml
Fizzadar: https://github.com/mautrix/meta/pull/196/commits/c907b93a612b507cffb94f1495bdcb649e67d1ce

pkg/connector/config.go
```diff
@@ -74,6 +82,8 @@ func upgradeConfig(helper up.Helper) {
 	helper.Copy(up.Bool, "send_presence_on_typing")
 	helper.Copy(up.Bool, "receive_instagram_typing_indicators")
 	helper.Copy(up.Bool, "disable_view_once")
+	helper.Copy(up.Bool, "thread_backfill", "enabled")
```
tulir:
```suggestion
	helper.Copy(up.Int, "thread_backfill", "batch_count")
```

pkg/connector/config.go
```diff
@@ -74,6 +82,8 @@ func upgradeConfig(helper up.Helper) {
...
+	helper.Copy(up.Str, "thread_backfill", "batch_delay")
```
tulir: I think duration parsing allows numbers too (mostly useful for `0`)

```suggestion
	helper.Copy(up.Str|up.Int, "thread_backfill", "batch_delay")
```

pkg/connector/threadbackfill.go
```diff
@@ -0,0 +1,89 @@
...
+
+	return m.runThreadBackfill(ctx)
+}
+
+func (m *MetaClient) runThreadBackfill(ctx context.Context) error {
+	log := zerolog.Ctx(ctx)
+	delay := m.Main.Config.ThreadBackfill.BatchDelay
+
+	batchCount := 0
+	lastMinThreadKey := m.Client.GetSyncGroupKeyStore(1).MinThreadKey
```
tulir: Can the key store be nil? (i.e. will this panic?)

pkg/connector/config.go
Fizzadar: https://github.com/mautrix/meta/pull/196/commits/bb63550e3fa6837b227c290a5a21e26dbe617b45

pkg/connector/config.go
Fizzadar: https://github.com/mautrix/meta/pull/196/commits/bb63550e3fa6837b227c290a5a21e26dbe617b45

pkg/connector/threadbackfill.go
Fizzadar: Don't need it at all I think: https://github.com/mautrix/meta/pull/196/commits/d6f6cf88c44d30bcc0e18f46c7ef679df7fb0aef

PR #201, opened 2026-01-16
client: allow reconnecting after a Disconnect call

Fizzadar:
Because the client is only recreated during a full reconnect, currently if you connect -> disconnect -> connect the second connect will immediately fail and set the bridge state to bad credentials.

### DIFF
--- a/pkg/connector/client.go
+++ b/pkg/connector/client.go
@@ -141,7 +141,7 @@ func (m *MetaConnector) getProxy(reason string) (string, error) {
 }
 
 func (m *MetaClient) ensureMessagixClient() {
-	if m.LoginMeta.Cookies != nil {
+	if m.LoginMeta.Cookies != nil && m.Client == nil {
 		m.LoginMeta.Cookies.Platform = m.LoginMeta.Platform
 		m.Client = messagix.NewClient(
 			m.LoginMeta.Cookies,


Fizzadar:
> It should never connect -> disconnect -> connect?

We do when handling push notifications via `ConnectBackground` which then calls disconnect and if we open the app we call connect again, since the user login is never replaced we don't re-create the client.

pkg/connector/client.go
```diff
@@ -555,8 +560,7 @@ func (m *MetaClient) FullReconnect() {
 	m.connectWaiter.Clear()
 	m.e2eeConnectWaiter.Clear()
 	m.disconnect(false)
-	m.Client = messagix.NewClient(m.LoginMeta.Cookies, m.UserLogin.Log.With().Str("component", "messagix").Logger(), m.Main.getMessagixConfig())
-	m.Client.SetEventHandler(m.handleMetaEvent)
+	m.ensureMessagixClient()
```
tulir: This call is redundant since Connect does it anyway

pkg/connector/client.go
```diff
@@ -100,9 +94,7 @@ func (m *MetaConnector) LoadUserLogin(ctx context.Context, login *bridgev2.UserL
 		igUserIDsReverse:  map[int64]string{},
 	}
 	c.editChannels = exsync.NewMap[string, chan *FBEditEvent]()
-	if messagixClient != nil {
-		messagixClient.SetEventHandler(c.handleMetaEvent)
-	}
+	c.ensureMessagixClient()
```
tulir: It might also make sense to drop this call so that a newly loaded client is in the same state as a disconnected one. Not sure if that would break something though

pkg/connector/client.go
Fizzadar: https://github.com/mautrix/meta/pull/201/commits/456437613c8437308175600eb9e0d09efe0a98a1

pkg/connector/client.go
Fizzadar: I think it looks safe => https://github.com/mautrix/meta/pull/201/commits/456437613c8437308175600eb9e0d09efe0a98a1

pkg/connector/client.go
```diff
@@ -149,6 +140,18 @@ func (m *MetaConnector) getProxy(reason string) (string, error) {
 	return respData.ProxyURL, nil
 }
 
+func (m *MetaClient) ensureMessagixClient() {
+	if m.LoginMeta.Cookies != nil {
```
tulir: This is called multiple times if the connection fails (e.g. due to network issues), so probably shouldn't keep recreating it

```suggestion
	if m.LoginMeta.Cookies != nil && m.Client == nil {
```

tulir (approved):

PR #215, opened 2026-01-30
handlemeta: insert messages received in `DeleteThenInsertMessages` txn

Fizzadar:


pkg/messagix/table/messages.go
```diff
@@ -558,6 +558,91 @@ func (ls *LSDeleteThenInsertMessage) GetThreadKey() int64 {
 	return ls.ThreadKey
 }
 
+func (ls *LSDeleteThenInsertMessage) ToWrappedMessage() *WrappedMessage {
```
Fizzadar, line 561: Horrible but 🤷 , we should figure out a way to do this without.. doing that

PR #218, opened 2026-02-06
handlemeta: fix IG typing portal key receiver

adamvy:


adamvy:
> hmm, could it just be inferred to be a group if threadID != userID? (as long as own typing notifications are ignored)

unsure, i dind't meant to send this as a ready to review pr was just experimenting with AI on solvin git

PR #232, opened 2026-02-26
connector,messagix: Messenger Lite push registration

radon-at-beeper:
This isn't really tested yet. Also, while I am refactoring the device ids etc to be saved on the login metadata, I don't think I'm restoring them to be used again when reconnecting an existing account; I think that is something I still need to fix.


pkg/connector/client.go
```diff
@@ -445,8 +445,9 @@ func (m *MetaClient) connectE2EE() error {
 		isNew = true
 		m.WADevice = m.Main.DeviceStore.NewDevice()
 	}
-	if suggested := m.Client.MessengerLite.GetSuggestedDeviceID(); suggested != uuid.Nil {
-		m.WADevice.FacebookUUID = suggested
+	if known := m.LoginMeta.MessengerLite.DeviceID; known != "" {
+		// we know this uuid is safe to parse because we generated it
+		m.WADevice.FacebookUUID = uuid.MustParse(known)
```
tulir, line 450: I think `uuid.UUID` is fine with being in a struct that gets json serialized, should probably just use the proper type there

radon-at-beeper: The thing about that is the default representation is lowercase, whereas Facebook wants to use uppercase for all the Messenger Lite stuff. I could use a wrapper type that uppercases during marshaling?

tulir: I only meant in the db metadata struct, you can probably stringify+uppercase when passing to facebook?

radon-at-beeper: Yeah, I was doing that before, but I felt like it was easier to treat the tokens opaquely since the encoding seemed like it could be important. If you prefer though I'm happy to go back to storing as underlying uuid.UUID and converting them on usage.

radon-at-beeper:
Putting on hold for now.

PR #243, opened 2026-03-07
handlemeta: handle instagram message request

rnons:


pkg/connector/events.go
```diff
@@ -569,6 +573,9 @@ func (r *FBChatResync) ShouldCreatePortal() bool {
 	if r.Raw == nil {
 		return false
 	}
+	if r.Info != nil && r.Info.MessageRequest != nil && *r.Info.MessageRequest {
```
Fizzadar, line 593: Should we check spam first? Is it possible to have a folder be a request _and_ marked as spam?

Fizzadar (changes requested):
One thing!

pkg/connector/events.go
```diff
@@ -656,3 +663,39 @@ func (r *FBChatResync) CheckNeedsBackfill(ctx context.Context, lastMessage *data
...
+func (evt *FBMessageRequestEvent) AddLogContext(c zerolog.Context) zerolog.Context {
+	return c.Int64("thread_id", evt.ThreadKey).Bool("message_request", true)
+}
+
+func (evt *FBMessageRequestEvent) GetSender() bridgev2.EventSender {
+	return bridgev2.EventSender{}
+}
+
+func (evt *FBMessageRequestEvent) GetChatInfo(ctx context.Context, portal *bridgev2.Portal) (*bridgev2.ChatInfo, error) {
+	info := evt.m.makeMinimalChatInfo(evt.ThreadKey, table.UNKNOWN_THREAD_TYPE)
```
tulir: Unknown thread type here sounds wrong. If we actually don't know the thread type, we can't safely create portals. The uncertain receiver method should also be implemented properly

pkg/connector/handlemeta.go
```diff
@@ -550,6 +568,11 @@ func markPortalAsEncrypted(ctx context.Context, portal *bridgev2.Portal) bool {
 	return false
 }
 
+func forcePortalBridgeInfoResync(ctx context.Context, portal *bridgev2.Portal) bool {
+	portal.UpdateBridgeInfo(ctx)
+	return false
+}
```
tulir: This doesn't seem like it should exist at all

tulir:
You mean if the bridge isn't online when the message request is accepted? Does the sync on non-cached reconnects not have enough data to automatically handle that? We should be getting 20 most recent chats automatically

Definitely can't spam requests for every chat on every full reconnect though

pkg/connector/events.go
```diff
@@ -112,7 +112,11 @@ func (evt *VerifyThreadExistsEvent) GetChatInfo(ctx context.Context, portal *bri
 			zerolog.Ctx(ctx).Trace().Any("response", resp).Msg("Requested full thread info")
 		}
 	}
-	return evt.m.makeMinimalChatInfo(evt.ThreadKey, evt.ThreadType, evt.ParentThreadKey), nil
+	chatInfo := evt.m.makeMinimalChatInfo(evt.ThreadKey, evt.ThreadType, evt.ParentThreadKey)
+	if evt.FolderName == folderPending {
+		chatInfo.MessageRequest = ptr.Ptr(true)
+	}
```
tulir: would this be reliable?

```suggestion
	chatInfo.MessageRequest = ptr.Ptr(evt.FolderName == folderPending)
```

tulir (commented):
This plus setting it in chat resync events would probably deal with full reconnects?

pkg/connector/client.go
```diff
@@ -358,6 +358,7 @@ func (m *MetaClient) connectWithTable(ctx context.Context, initialTable *table.L
 	m.UserLogin.RemoteProfile.Avatar = m.Ghost.AvatarMXC
 
 	m.initialTable.Store(initialTable)
+	m.startMessageRequestSync(ctx)
```
tulir: Is the extra call required? If message requests are already pushed while online and included in the 20 recent chats that are in the initial table after full reconnects, then I don't think we really need the extra request.

Could maybe be made configurable to sync everything?

tulir (approved):
Seems like a bunch of merge conflicts appeared too. Probably fine with or without the extra request

pkg/connector/client.go
rnons: updated, you're right, removed the extra call

Fizzadar (approved):
Sorry about the delay @rnons!

PR #264, opened 2026-05-05
messagix/bloks: support bk.action.core.Let

iFixRobots:
Adds the missing `bk.action.core.Let` opcode. Without it, the Messenger Lite login flow fails on the email/password screen with `tapping login button: on_touch_down: unimplemented function bk.action.core.Let (2 args)` — Facebook's login script started using `Let` to lazy-init per-screen animation state.

`Let(currentOrNull, FuncConst(initializer))`: if `currentOrNull` evaluates to a non-null value (a previously cached value), return it; otherwise apply the FuncConst lambda in arg 1 and return its result. Implementation matches the existing `Coalesce` pattern plus the `Apply` lambda invocation.

Reproduced and verified against a real captured bundle: the on_touch_down handler now executes through to `com.bloks.www.bloks.caa.login.async.send_login_request` cleanly via `pkg/messagix/bloks/cmd -file <bundle> -action -login`.

Closes [PLAT-36280](https://linear.app/beeper/issue/PLAT-36280).

### DIFF
--- a/pkg/messagix/bloks/interp.go
+++ b/pkg/messagix/bloks/interp.go
@@ -368,6 +368,23 @@ func (i *Interpreter) Evaluate(ctx context.Context, form *BloksScriptNode) (*Blo
 			return first, nil
 		}
 		return i.Evaluate(ctx, &call.Args[1])
+	case "bk.action.core.Let":
+		// Lazy-init: return arg0 if non-null, else apply the FuncConst in arg1.
+		current, err := i.Evaluate(ctx, &call.Args[0])
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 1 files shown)


radon-at-beeper:
Could also implement this as a macroexpansion since the code as written is basically saying "do `bk.action.bool.Or` then `bk.action.core.Apply`". Example:

https://github.com/mautrix/meta/blob/cd0293c0e4ea97ca922070be38eefb201c720069/pkg/messagix/bloks/interp.go#L800-L815

But that's not really necessary, so approving as is.

radon-at-beeper (approved):

PR #270, opened 2026-06-01
messagix/bloks: Allow logging redacted Bloks payloads

radon-at-beeper:
* New config option `network.log_redacted_bloks_payloads`. When enabled, log the Bloks payloads at debug level rather than just trace, but do it in redacted form.
* New framework for doing the redaction. The way it works is a recursive `.Redact()` method that you call to replace all sensitive identifiers in-place, then you `.Print()` as before.
* Update `.Print()` to take an `io.Writer` so that printing to a string is possible.
* Renamed `BloksJavascriptValue` to capitalize properly `BloksJavaScriptValue`.
* Fixed a bug where JS literals were printed with some bogus fields generated by Golang, because we weren't calling `.Flatten()` properly before marshaling them.
* New `-redact` option to the cmdline tool, to test the redaction framework easily. Analogous to `-print`. Accept `-` as a value for `-filename`, to read from stdin per UNIX conventions.
* Replaced `BloksJavaScriptValue` with new type alias `BloksScriptLiteralValue` as the field type in `BloksScriptLiteral`. Previously we were reusing `BloksJavaScriptValue` between both the script literal context and the actual JS literal context, despite their having materially different internal formats within the `any`, and converter functions for going between them. Now the typing makes it clear which internal structure to expect depending on context.

Redaction framework:
* Heuristics based on Bloks scripting. Certain functions are only ever passed nonsensitive data in certain argument positions. If those arguments are string literals, they're left alone.
* Heuristics based on known-safe patterns within string literals, for example ones that start with `com.bloks` or static CDN URLs.
* Heuristics for detecting what are likely to be English sentences or coding identifiers, based on approximate symbol composition rather than a heavier dictionary/entropy-based approach. Works very well in practice.
* Replace redacted strings with sha256 hash of their contents, and include length+formatting information.
* Redact long integers by zeroing out everything past the first 10 significant digits, to avoid capturing credentials or tokens represented as integers.
* Traverse into string literals that appear to contain stringified JSON objects, to redact more precisely. Due to Golang limitations, this can reorder field names.

Design decision: redaction framework applies _after_ parsing and unminification, because that's when we have rich typing information available to inform the redaction. There's no way to re-minify, and formatting won't be preserved perfectly, unlike the existing payload trace logs which are dumped before parsing.

Design decision: redaction framework operates in-place, and there's no way to make a copy. What you do is just parse the raw payload string twice and then redact one of them. The implementation is way simpler this way and the performance implication is really not that significant, plus this is only a debugging feature.

Design decision: gzip and base64-encode redacted payload before logging. Some payloads can be as large as 1MB, totally inappropriate for a log line. gzip compression reduces that to 50KB, which is still bad but is a lot better.

### DIFF
--- a/pkg/connector/client.go
+++ b/pkg/connector/client.go
@@ -75,8 +75,9 @@ type MetaClient struct {
 
 func (m *MetaConnector) getMessagixConfig() *messagix.Config {
 	return &messagix.Config{
-		MayConnectToDGW: m.Config.ReceiveInstagramTypingIndicators,
-		ClientSettings:  m.Bridge.GetHTTPClientSettings(),
+		MayConnectToDGW:          m.Config.ReceiveInstagramTypingIndicators,
+		ClientSettings:           m.Bridge.GetHTTPClientSettings(),
+		LogRedactedBloksPayloads: m.Config.LogRedactedBloksPayloads,
 	}
 }
 
--- a/pkg/connector/config.go
+++ b/pkg/connector/config.go
@@ -45,6 +45,7 @@ type Config struct {
 	ReceiveInstagramTypingIndicators bool `yaml:"receive_instagram_typing_indicators"`
 	DisableViewOnce                  bool `yaml:"disable_view_once"`
 	MarketplaceSpace                 bool `yaml:"marketplace_space"`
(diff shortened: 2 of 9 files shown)


Fizzadar (approved):
Nice, LGTM!

PR #276, opened 2026-06-11
handlemeta: mark as msg request only for pending folder

rnons:


tulir (approved):
Looks fine, hopefully it's correct

PR #287, opened 2026-06-16
connector: backfill DM peer UserInfo on resync

Fizzadar:


tulir:
Hmm, this looks wrong too because `makeMinimalChatInfo`  already adds member entries for DMs. Is the problem that `FBChatResync.Members` contains yourself, but not the other user?

Fizzadar:
> Hmm, this looks wrong too because `makeMinimalChatInfo` already adds member entries for DMs. Is the problem that `FBChatResync.Members` contains yourself, but not the other user?

Seems like neither are set and we fallthrough to `makeMinimalChatInfo`  but that doesn't include user info, so the room is re-created with no profile on the other user. We also don't have any ghost for mautrix to pull the profile from.

I have not reproduced this directly, however.

tulir:
User info is fetched separately if needed. It's only included in the member list if conveniently available

PR #293, opened 2026-06-21
handlematrix: support facebook message requests

rnons:


pkg/connector/handlemeta.go
```diff
@@ -416,10 +416,15 @@ func (m *MetaClient) parseTable(ctx context.Context, tbl *table.LSTable) (innerQ
 		innerQueue = append(innerQueue, rs)
 	}
 	for _, thread := range tbl.LSDeleteThenInsertThread {
+		waKey := thread.ThreadKey
```
tulir: This variable name and all the comments are the wrong way around: the original key is the facebook key, the mapped one is the whatsapp key (jid)

tulir (approved):

PR #296, opened 2026-07-06
messagix/bloks: improve transient token error message

Fizzadar:
In most cases retrying seems to work since this is after the approval bit, we're just missing the exchange of token to cookies atm.

### DIFF
--- a/pkg/messagix/bloks/selenium.go
+++ b/pkg/messagix/bloks/selenium.go
@@ -1074,13 +1074,16 @@ func (b *Browser) DoLoginStep(ctx context.Context, userInput map[string]string) 
 		methodNames := []string{}
 
 		knownMethods := map[string]bool{
+			"Authentication app": true,
+			"Email":              true,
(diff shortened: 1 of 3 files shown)


pkg/messagix/bloks/selenium.go
```diff
@@ -1074,13 +1074,16 @@ func (b *Browser) DoLoginStep(ctx context.Context, userInput map[string]string)
...
 		knownMethods := map[string]bool{
+			"Authentication app": true,
+			"Email":              true,
+			"Text message":       true,
+			"Backup code":        true,
+			"WhatsApp":           true,
+			// Disabled because there's no reasonable way to support this outside of a webview
+			"Verify with Google": false,
+			// Intentionally last because this tends to give transient_token results rather than
+			// cookies directly, which requires a currently unknown extra step.
```
radon-at-beeper, line 1085: Just FYI this definitely doesn't do what you want, because Golang maps are unordered.

PR #297, opened 2026-07-09
PLAT-37849 Use share ActionUrl as Instagram media-share caption URL

Fizzadar:
Instagram carousel-child shares were captioned with the parent post URL synthesized from the shortcode, so the caption link opened a different item than the embedded video. This prefers the cleaned `CTA.ActionUrl` as the external/caption URL when it verifiably points at an instagram.com `/p/`, `/reel/`, or `/reels/` path (absolute or relative forms), falling back to the synthesized URL otherwise. Strict host/path guards mean the worst case is unchanged from today.

### DIFF
--- a/pkg/msgconv/from-meta.go
+++ b/pkg/msgconv/from-meta.go
@@ -508,6 +508,10 @@ func removeLPHP(addr string) string {
 	return addr
 }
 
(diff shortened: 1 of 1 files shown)


Fizzadar:
New IG changes make this irrelevant!

PR #298, opened 2026-07-13
messagix/bloks: Handle passkey screen

radon-at-beeper:
For now, just cancel out of it and go back to the regular MFA screen.


radon-at-beeper:
Confirmed this works.

PR #300, opened 2026-07-14
messagix/bloks: error handling fixes

Fizzadar:


pkg/messagix/bloks/selenium.go
```diff
@@ -741,6 +739,9 @@ func (b *Browser) DoLoginStep(ctx context.Context, userInput map[string]string)
 		if !definitelyNotPhoneNumberRegexp.MatchString(username) {
 			return nil, ErrLoginPhoneNumber
 		}
+		if strings.Contains(username, ":") { // covers MXIDs
+			return nil, ErrLoginInvalidUsername
+		}
```
radon-at-beeper, line 747: Wait, why would somebody ever try to enter a federated Matrix account ID to... Facebook?

radon-at-beeper (approved):

pkg/messagix/bloks/selenium.go
Fizzadar: Good question! This is a failing of the UX I think, not clear enough with what we're expecting or that this is a bridge login flow. Proper fix is in the clients to make that clearer.

PR #301, opened 2026-07-14
messagix/bloks: implement timer cancel function

Fizzadar:
Based on bloks debug payload, we see the timer get cancelled and immediately restarted, without this change we error at the cancel. Bloks debug payload:

```
TakeLast(
  If( is_timer_running AND approval_id != "XXX",   ← this page's approval id
      { timer.Cancel(GetArg 0, "approve_from_another_device_polling_timer")
        is_timer_running := false },                             ← kill a STALE timer
      null ),
  If( is_timer_running,
      null,                                                      ← already running: no-op
      { is_timer_running := true
        timer.Start(GetArg 0, 2000, true,                        ← restart: poll every 2000ms
          λ: If(is_loading, null,
                invoke "com.bloks.www.two_step_verification.afad_state.async" ...),
          "approve_from_another_device_polling_timer") } )
)
```


pkg/messagix/bloks/selenium.go
```diff
@@ -1255,9 +1266,18 @@ func (b *Browser) DoLoginStep(ctx context.Context, userInput map[string]string)
...
 		for b.State == StateAFADPageWaiting {
+			if b.AFADCallback == nil {
+				return nil, ErrLoginAFADStopped
+			}
 			time.Sleep(b.AFADInterval)
 			err := b.AFADCallback()
 			if err != nil {
+				if ctxErr := ctx.Err(); ctxErr != nil {
+					if errors.Is(ctxErr, context.DeadlineExceeded) {
+						return nil, ErrLoginAFADTimeout
```
radon-at-beeper: This seems wrong. If there is a context error then that means the user is gone already, but you're trying to return an error that is intended to display to the user.

radon-at-beeper (approved):

pkg/messagix/bloks/selenium.go
Fizzadar: Fixed, in testing against the debug payloads I also spotted that every cancel function also displayed screen `com.bloks.www.approve_from_another_device.xmds.challenged_device_denied` so I've added support for that in here too.

radon-at-beeper (approved):

PR #303, opened 2026-07-17
messagix/bloks: support single continue MFA button

Fizzadar:
Got to say, the framework/payload debugging makes this an excellent dev experience.

- fix panic on the nil
- same user also hit a path where MFA is a single continue/cancel button, so handle that too


radon-at-beeper (approved):

radon-at-beeper:
This is stale as I by now have implemented all of the relevant changes separately. Also the implementation for ar.auth_method is most likely wrong, we should be delegating to the method picker instead of continuing to a preselected method. I've implemented that.

PR #305, opened 2026-07-21
handle*: support encrypted message requests properly

rnons:


pkg/messagix/lightspeed/decode_test.go
tulir, line 1: This test looks fairly useless

pkg/connector/handlemeta.go
```diff
@@ -801,6 +841,9 @@ func (tm threadMaps) MapWhatsAppThreadKey(fbKey int64) int64 {
 	if waKey, ok := tm.waThreadMap[fbKey]; ok {
 		return waKey
 	}
+	if waKey, ok := tm.m.hybridThreadMap.Get(fbKey); ok {
+		return waKey
+	}
```
tulir: Is this map actually needed? If it's only for LSDeletePartialThread, just let it fall back to the db instead of maintaining a separate cache

pkg/connector/events.go
```diff
@@ -281,7 +281,15 @@ func (evt *EnsureWAChatStateEvent) GetChatInfo(ctx context.Context, portal *brid
...
 		}
-		return evt.m.makeWADirectChatInfo(evt.JID), nil
+		info := evt.m.makeWADirectChatInfo(evt.JID)
+		fbThreadKey, messageRequest, err := evt.m.Main.DB.GetHybridThreadInfoByJID(ctx, evt.m.UserLogin.ID, metaid.ParseFBPortalID(evt.GetPortalKey().ID))
+		if err != nil {
+			zerolog.Ctx(ctx).Warn().Err(err).Msg("Failed to restore hybrid thread info")
+		} else if fbThreadKey != 0 {
+			info.MessageRequest = ptr.Ptr(messageRequest)
+			info.ExtraUpdates = bridgev2.MergeExtraUpdaters(info.ExtraUpdates, setFBThreadKey(fbThreadKey))
+		}
```
tulir: This looks like it should be inside makeWADirectChatInfo?

pkg/connector/capabilities.go
```diff
@@ -229,12 +237,19 @@ func init() {
...
 func (m *MetaClient) GetCapabilities(ctx context.Context, portal *bridgev2.Portal) *event.RoomFeatures {
-	switch portal.Metadata.(*metaid.PortalMetadata).ThreadType {
+	meta := portal.Metadata.(*metaid.PortalMetadata)
+	switch meta.ThreadType {
 	case table.COMMUNITY_GROUP:
 		return metaCapsWithThreads
 	case table.ENCRYPTED_OVER_WA_ONE_TO_ONE:
+		if meta.FBThreadKey != 0 {
+			return metaCapsWithE2EDelete
+		}
```
tulir: Capabilities don't need to differentiate this unless there are legitimate cases where we won't have the FB thread key

pkg/connector/capabilities.go
```diff
@@ -55,7 +55,7 @@ func (m *MetaConnector) GetCapabilities() *bridgev2.NetworkGeneralCapabilities {
 }
 
 func (m *MetaConnector) GetBridgeInfoVersion() (info, caps int) {
-	return 1, 15
+	return 1, 16
```
tulir, line 58: The cap ID should also be updated to actually have it resend

pkg/metadb/00-latest-schema.sql
```diff
@@ -1,4 +1,4 @@
--- v0 -> v11 (compatible with v11+): Latest schema
+-- v0 -> v12 (compatible with v12+): Latest schema
```
tulir: This looks backwards compatible

```suggestion
-- v0 -> v12 (compatible with v11+): Latest schema
```

tulir (approved):

PR #308, opened 2026-07-27
connector: don't save reconnection state before events are handled

Fizzadar:
The messagix sync cursor advances as soon as a table is parsed and queued, so persisting the dumped state while the derived events are still queued in memory. If the process restarts in this situation those events are lost forever.

### DIFF
--- a/pkg/connector/client.go
+++ b/pkg/connector/client.go
@@ -40,6 +40,7 @@ type MetaClient struct {
 	initialTable        atomic.Pointer[table.LSTable]
 	initialTableHandled atomic.Bool
 	parsedTables        chan *parsedTable
+	pendingTables       atomic.Pointer[atomic.Int64]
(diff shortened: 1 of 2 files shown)


pkg/connector/client.go
```diff
@@ -500,6 +506,12 @@ func (m *MetaClient) saveConnectionState(ctx context.Context, state json.RawMess
 	if !m.Main.Config.CacheConnectionState {
 		return
 	}
+	if pending := m.pendingTables.Load().Load(); pending != 0 {
+		zerolog.Ctx(ctx).Warn().
+			Int64("pending_tables", pending).
+			Msg("Not saving reconnection state, some events haven't been handled yet")
```
Fizzadar, line 512: I think this is safe as we'll just save it later?

tulir:
Hmm, the saving should probably just be a part of the event stream like it is on instagram rather than a hacky atomic atomic integer

PR #310, opened 2026-07-29
messagix: Implement transient token exchange

Fizzadar:
Uses newly discovered endpoints and includes debugging facilities to make it possible to test on any account, even those that don't hit this issue.


pkg/connector/config.go
```diff
@@ -48,6 +48,7 @@ type Config struct {
 	DisableViewOnce                  bool `yaml:"disable_view_once"`
 	MarketplaceSpace                 bool `yaml:"marketplace_space"`
 	LogRedactedBloksPayloads         bool `yaml:"log_redacted_bloks_payloads"`
+	DumpMessengerLiteLoginTokens     bool `yaml:"dump_messenger_lite_login_tokens"`
```
Fizzadar: Not sure we want to include this at all, once confirmed working.

Fizzadar:
Ready for review but pending deciding if we keep the debug flag.

PR #316, opened 2026-08-11
instameow: Native login for instagram

highesttt:
### Checklist

PLAT-38268


pkg/messagix/bloks/interp.go
```diff
@@ -1260,6 +1697,11 @@ func (i *Interpreter) Evaluate(ctx context.Context, form *BloksScriptNode) (*Blo
 			return nil, err
 		}
 		return nil, fmt.Errorf("%s", msg)
+	case "bk.action.io.CurrentTimeMillis":
+		// Android's System.currentTimeMillis() returns a signed 64-bit count of
+		// milliseconds since the Unix epoch. Keep that representation so Bloks
+		// timestamp arithmetic doesn't lose precision.
```
tulir: completely useless comments should be avoided

```suggestion
```

pkg/messagix/bloks/interp.go
```diff
@@ -1271,6 +1713,35 @@ func (i *Interpreter) Evaluate(ctx context.Context, form *BloksScriptNode) (*Blo
...
+		// This can occur in the unused animation branch paired with
+		// IsInitialized. Return the stable pre-animation value.
+		return BloksLiteralOf(float64(0)), nil
+	case "bk.action.animated.Create",
+		"bk.action.animated.Parallel",
+		"bk.action.animated.easing.CreateCubicBezier",
+		"bk.action.template.Make",
+		"bk.action.bloks.Find",
+		"bk.action.ig.identitysafety.livechat.GetStartChatParams",
+		"bk.action.context.Get":
```
tulir: These should be merged with the BloksNull return below

pkg/messagix/bloks/minify.json
```diff
@@ -1679,7 +1680,9 @@
     "jbv": "bk.action.caa.ShowLoggedInResetPasswordV2",
     "jcq": "bk.action.oxygen.TriggerSync",
     "jfy": "bk.versioning.bloks.SupportsPureBloksBANI",
-    "jix": "bk.action.io.CurrentTimeMillisV2"
+    "jix": "bk.action.io.CurrentTimeMillisV2",
+    "jmu": "bk.action.num.Add",
```
tulir: This looks like a merge conflict with https://github.com/mautrix/meta/commit/1821914c251b26523524fa91850f746cc065ad8a

pkg/messagix/crypto/password.go
```diff
@@ -65,14 +83,72 @@ func EncryptPassword(platform types.Platform, pubKeyId int, pubKey, password str
...
+	block, err := aes.NewCipher(sessionKey)
+	if err != nil {
+		return "", ErrAESCreation
+	}
+	aead, err := cipher.NewGCM(block)
+	if err != nil {
+		return "", ErrGCMCreation
+	}
+	iv := make([]byte, aead.NonceSize())
+	if _, err = rand.Read(iv); err != nil {
```
tulir: Should always use `random.Bytes` instead of manual reads

pkg/messagix/httpclient/http.go
tulir, line 1: Most changes in this file seem unnecessary and should probably be reverted

pkg/messagix/httpclient/http_test.go
tulir, line 1: This file is probably not needed either

pkg/instameow/login_web.go
```diff
@@ -0,0 +1,126 @@
...
+		cookies.IGCookieDeviceID:  c.cookies.Get(cookies.IGCookieDeviceID),
+	})
+	c.cookies.IGWWWClaim = ""
+
+	c.configs = httpclient.NewConfigs(c)
+	c.http.SetConfigs(c.configs)
+	moduleLoader := httpclient.NewModuleParser(c, c.http, c.configs)
+	moduleLoader.LS = nil
+	baseURL := c.GetEndpoint("base_url")
+	loginPageURL := baseURL + "/accounts/login/"
```
tulir: This should be in the endpoint map rather than manually appending

pkg/metadb/00-latest-schema.sql
```diff
@@ -89,3 +89,12 @@ CREATE TABLE meta_hybrid_thread (
     CONSTRAINT meta_hybrid_thread_user_login_fkey FOREIGN KEY (bridge_id, login_id)
         REFERENCES user_login (bridge_id, id) ON DELETE CASCADE ON UPDATE CASCADE
 );
+
+CREATE TABLE meta_instagram_login_device (
+    bridge_id         TEXT NOT NULL PRIMARY KEY,
+    phone_id          TEXT NOT NULL,
+    device_id         TEXT NOT NULL,
```
tulir: This table has to be scoped to user, global is wrong. I assume we want to use the same id for all logins of a user? (if not, then the table shouldn't exist at all and it should just be user login metadata, which can be reused for relogins to the same account)

pkg/metadb/00-latest-schema.sql
```diff
@@ -89,3 +89,12 @@ CREATE TABLE meta_hybrid_thread (
...
+    advertising_id    TEXT NOT NULL,
+    android_device_id TEXT NOT NULL,
+    machine_id        TEXT NOT NULL DEFAULT ''
```
tulir: Also no point in default values when a non-null value is always provided

pkg/instameow/login_web.go
```diff
@@ -0,0 +1,126 @@
...
+		if parseErr == nil && result.Message != "" {
+			return fmt.Errorf("instagram web login failed: %s", result.Message)
+		}
+		return fmt.Errorf("instagram web login request failed: %w", requestErr)
+	} else if response == nil {
+		return errors.New("instagram web login returned no response")
+	} else if parseErr != nil {
+		return fmt.Errorf("failed to parse Instagram web login: %w", parseErr)
+	} else if result.TwoFactorRequired {
+		return errors.New("instagram web session requires an additional two-factor login")
```
tulir: Is the idea that the web login api doesn't require 2fa when mobile cookies are provided or something?

pkg/messagix/bloks/bloks.go
```diff
@@ -14,12 +14,17 @@ const BloksVersionIOS = "e74cbfe42b6152a793f3ee3d81eeba4c48b5c4477b54d05ab8d8bce
...
+const BloksVersionInstagram = "3ae1e6445cdefeda05bde09f57499d84157ad74699044bf7fe9ca9183150f17a"
+
 func GetBloksVersion(p types.Platform) (string, error) {
 	switch p {
 	case types.MessengerLiteIOS:
 		return BloksVersionIOS, nil
 	case types.MessengerLiteAndroid:
 		return BloksVersionAndroid, nil
+	case types.Instagram:
+		return BloksVersionInstagram, nil
```
radon-at-beeper: InstagramAndroid, yes?

pkg/messagix/bloks/interp.go
```diff
@@ -584,6 +820,72 @@ func (i *Interpreter) Evaluate(ctx context.Context, form *BloksScriptNode) (*Blo
...
+			sum := firstInt + secondInt
+			if (secondInt >= 0 && sum >= firstInt) || (secondInt < 0 && sum < firstInt) {
+				return BloksLiteralOf(sum), nil
+			}
+		}
+		sum := firstFloat + secondFloat
+		if sum >= -2147483648 && sum <= 2147483647 && float64(int64(sum)) == sum {
+			return BloksLiteralOf(int64(sum)), nil
+		}
+		return BloksLiteralOf(sum), nil
```
radon-at-beeper: needs to be merged with https://github.com/mautrix/meta/pull/315 probably

pkg/messagix/bloks/interp.go
```diff
@@ -984,6 +1322,78 @@ func (i *Interpreter) Evaluate(ctx context.Context, form *BloksScriptNode) (*Blo
...
 		}
 		return BloksLiteralOf(string(encoded)), nil
+	case "bk.action.string.Concat":
+		parts := make([]string, 0, len(call.Args))
+		for idx := range call.Args {
+			value, err := i.Evaluate(ctx, &call.Args[idx])
+			if err != nil {
+				return nil, err
+			}
+			if array, ok := value.Value().([]*BloksScriptLiteral); ok && len(call.Args) == 1 {
```
radon-at-beeper: why not using `evalAs[]`?

pkg/messagix/bloks/interp_test.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+func TestGenericNumberAdd(t *testing.T) {
+	unminifier, err := GetUnminifier(&BloksBundle{})
+	if err != nil {
+		t.Fatalf("failed to load Bloks unminifier: %v", err)
+	}
+	const function = BloksFunctionID("bk.action.num.Add")
+	if got := unminifier.Functions["jmu"]; got != function {
+		t.Fatalf("jmu unminified to %q instead of %q", got, function)
```
radon-at-beeper: why are we testing this for every function...?

pkg/messagix/bloks/selenium.go
```diff
@@ -335,19 +341,196 @@ type Browser struct {
...
+		"Meta can't send an SMS code right now. Try again later or use a different two-factor method",
+	)
+}
+
+func (b *Browser) mandatoryPasskeyLoginError() bridgev2.RespError {
+	return b.platformLoginError(
+		ErrLoginMandatoryPasskey,
+		"Meta is requiring a passkey, which is not supported. Please add a different two-factor method in the official app or website",
+	)
+}
```
radon-at-beeper: I feel like we already had a list of these RespError constants, this is introducing a second list in a different place.

pkg/messagix/bloks/selenium.go
```diff
@@ -624,7 +930,7 @@ func NewBrowser(cfg *BrowserConfig) (*Browser, error) {
 			case "com.bloks.www.two_step_verification.no_op_captcha":
 				newState = StateSilentCaptchaPage
 			case "com.bloks.www.two_step_verification.google_recaptcha":
-				return ErrLoginReCaptcha
+				return b.reCaptchaLoginError()
```
radon-at-beeper: Why change to a function?

pkg/messagix/bloks/selenium_test.go
```diff
@@ -0,0 +1,917 @@
+package bloks
```
Fizzadar: IMO we drop this in favour of proper tests against captured payloads which bloks helpfully does already, that's a much more complete test and makes unit tests redundant.

pkg/messagix/bloks/selenium.go
Fizzadar: Funny you mention, we have them duplicated 3 times atm (searching for FI.MAU.META_MISSING_COOKIES). Login, iglogin and bloks all have that error code.

@highesttt can we just reuse those or better clean them all up and dedup.

pkg/messagix/bloks/selenium.go
```diff
@@ -579,11 +837,48 @@ func NewBrowser(cfg *BrowserConfig) (*Browser, error) {
 			switch name {
 			case "com.bloks.www.caa.login.login_homepage":
 				newState = StateEmailPasswordPage
-			case "com.bloks.www.caa.ar.code_entry",
-				"com.bloks.www.ap.two_step_verification.code_entry":
+			case "com.bloks.www.caa.login.landing_screen":
+				if !b.Config.Platform.IsInstagram() {
```
Fizzadar: Is there any benefit to locking these to IG only? If there's a chance they might be used elsewhere...

pkg/messagix/bloks/selenium.go
```diff
@@ -797,37 +1258,37 @@ func (b *Browser) DoLoginStep(ctx context.Context, userInput map[string]string)
 		b.State = StateInitial
 
 	case StateInitial:
+		if b.Config.Platform.IsInstagram() {
```
Fizzadar: Why reuse StateInitial here? Would it not be cleaner with a StateInitialMessenger, StateInitialInstagram?

Fizzadar (changes requested):
Left some thoughts, my main worry is the bloks/selenium code seems like everything switches depending on ig or fb, in which case why share the struct in the first place?

pkg/instameow/login_web.go
```diff
@@ -0,0 +1,640 @@
+// mautrix-meta - A Matrix-Facebook Messenger and Instagram DM puppeting bridge.
```
Fizzadar, line 1: How does this interact with `login_web.go?` I don't quite follow the differences between them...

pkg/instameow/login_web.go
```diff
@@ -0,0 +1,640 @@
...
+		TOTP:     info.TOTP,
+		SMS:      info.SMS,
+		WhatsApp: info.WhatsApp,
+	}, nil
+}
+
+// CreateInstagramWebSession turns a completed native login into a web session
+// that can be used by Instagram's web messaging APIs and realtime stream. A
+// returned challenge must be completed with CompleteInstagramWebSessionTwoFactor.
+func (c *Client) CreateInstagramWebSession(
```
Fizzadar, line 313: Ah so re:my comment above - this file is all about turning the mobile session into a web one?

And assuming that - what is it the mobile API gives us that this web API cannot beforehand?

Fizzadar (commented):
Couple of questions, still need to review all the bloks changes.

pkg/instameow/socket.go
```diff
@@ -150,7 +150,7 @@ func (c *Client) getSocketOptions() dgw.SocketOptions {
 		Log:        c.log.With().Str("socket", "main").Logger(),
 		Facebook:   false,
 		AppID:      c.configs.BrowserConfigTable.DGWWebConfig.AppID,
-		UserID:     c.configs.BrowserConfigTable.PolarisViewer.ID,
+		UserID:     c.configs.BrowserConfigTable.PolarisViewer.Data.Fbid,
```
Fizzadar, line 153: I believe this will break existing cookie based logins? Have we tested that?

pkg/igconnector/login_native.go
```diff
@@ -0,0 +1,331 @@
...
+	if err != nil {
+		return nil, err
+	}
+	loginTransport := m.transport
+	m.transport = nil
+	if loginTransport != nil {
+		originalTransport := client.GetHTTP().HTTP.Transport
+		client.GetHTTP().HTTP.Transport = loginTransport
+		defer func() {
+			client.GetHTTP().HTTP.Transport = originalTransport
```
Fizzadar: This will race against the client that re-uses it after loginWithCookies completes. Better to pass a callback or something to that and explicitly clear the transport before starting the main bridge client.

Fizzadar (changes requested):
Two more things I spotted plus my request on matrix to test a messenger login with these changes to confirm nothing broke there (changes look like that is safe to me).

pkg/instameow/socket.go
tulir: I think it's actually correct for all logins, the old user id was wrong for this context (but the server didn't care)

highesttt: Login with cookies works fine with this change, messages work fine too

Fizzadar (approved):

PR #324, opened 2026-08-21
push: fix reload-page recovery when registering pusher

Fizzadar:
Pass back proper resp err if valid from push registration (seen in the wild the "page needs reload" error).


tulir (approved):
Looks about right

PR #325, opened 2026-08-21
connector: retry after consent required error

Fizzadar:
I have observed multiple occurrences of this where bridges land in this consent required error but reconnecting/reloading the page fixes it.

I guess this is fine because it'll just fall back into another bad creds if the user is indeed blocked, but feels like there's something else going on here 🤔


Fizzadar:
Too risky! We'll have the request bubble up in the client.

PR #336, opened 2026-08-28
login: reverted login_web.go and implemented native login via web

highesttt:
Changelog:
- reverted `login_web` deletion
- handles accounts with multiple instagram profiles (login_account_manager.go)
- replaced the mobile login entrypoint with web (login_native)

Tested on cloud and selfhost, on an account with 2FA & Account Manager with both Profiles

### Checklist

### DIFF
--- a/pkg/igconnector/login.go
+++ b/pkg/igconnector/login.go
@@ -57,7 +57,7 @@ var (
 )
 
 func (ic *IGConnector) GetLoginFlows() []bridgev2.LoginFlow {
-	return []bridgev2.LoginFlow{loginFlowInstagramPassword, loginFlowInstagram}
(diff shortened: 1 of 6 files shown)


pkg/igconnector/login.go
```diff
@@ -57,7 +57,7 @@ var (
 )
 
 func (ic *IGConnector) GetLoginFlows() []bridgev2.LoginFlow {
-	return []bridgev2.LoginFlow{loginFlowInstagram, loginFlowInstagramPassword}
+	return []bridgev2.LoginFlow{loginFlowInstagramPassword, loginFlowInstagram}
```
tulir: The order shouldn't be changed anymore

```suggestion
	return []bridgev2.LoginFlow{loginFlowInstagram, loginFlowInstagramPassword}
```

Fizzadar (approved):

Issue #338, opened 2026-08-29
Instagram generic_xma postback buttons are hidden from GraphQL


giuseppebisemi:
Creator bots (Manychat and similar comment-to-DM flows) send DM cards with tappable buttons. Instagram iOS renders those buttons. GraphQL usually returns `placeholder`, so the bridge emits the unsupported-message stub. Instagram web does the same, so this is not a Matrix client bug.

I fetched the same thread with GET `/api/v1/direct_v2/threads/{thread_id}/` on `www.instagram.com`, using the existing `useragent.AndroidUserAgent` and `X-IG-App-ID: 567067343352427`. That returns `item_type=generic_xma` and a `cta_buttons` array. Two shapes:

- `cta_type=xma_web_url` plus `action_url`. These can be HTML links. GraphQL already parses `cta_buttons` on XMA; if `action_url` is set, `wrapXMACaption` can emit an `<a>`. That change is small and I can send it as its own PR.

- `cta_type=postback`, empty `action_url`, payload of the form `1:GT-PB:ACT::<hex>`. Typing the button title as a normal DM does not run the bot. Manychat waits for Instagram's `messaging_postbacks` webhook.

An iOS tap later shows up in Direct as text plus `send_attribution=business_generic_template`. After the fact, that stored item looks a lot like `IGDirectTextSendMutation` with the same attribution. Manychat still ignores the GraphQL send. The webhook cares about the write path, not the fields sitting on the item afterwards.

POST `/api/v1/direct_v2/threads/broadcast/text/` with web cookies redirected to `/accounts/login/` and cleared `sessionid`. GET with the Android UA did not. I would not call that POST from a cookie login.

The send that did fire `messaging_postbacks` went over the existing mqttbypass socket, topic `/ig_send_message` (same path typing uses):

```json
{
  "action": "send_item",
  "item_type": "text",
  "text": "<button title>",
  "send_attribution": "business_generic_template",
  "cta_type": "postback",
  "postback_payload": "1:GT-PB:ACT::<hex>"
}
```

plus `thread_id` / `client_context` as usual.

I have a local patch that GETs the thread (Android UA) to recover the buttons, then sends taps over MQTT. It is a workaround: Android UA on a web session, and an undocumented MQTT payload. I would understand if that stayed out of tree. Native button chips in the Matrix client are a client problem either way.

If you want the GET+MQTT patch I can open a PR. If not, this issue is the writeup.

PR #339, opened 2026-08-29
msgconv/xma: render GraphQL CTA buttons with action_url as links

giuseppebisemi:
## Summary
- Turn GraphQL XMA `cta_buttons` that already have `action_url` into HTML links in the caption.
- Skip buttons with no URL (postbacks). Those are a different problem: #338.

## Test plan
- [x] `go test ./pkg/msgconv/igconv/`
- [x] Unit test: a CTA with `action_url` becomes an `<a href>`
- [x] Unit test: a CTA with only a title and no URL is omitted


giuseppebisemi:
Closing for now. The protocol notes stay in #338.

PR #340, opened 2026-08-31
messagix/bloks: allow choosing a different mfa type

radon-at-beeper:
This implements the interface defined in https://github.com/mautrix/go/pull/560 to allow canceling the MFA step after selecting an MFA method, and having that return the user to the MFA selection screen to choose a different method. The option to go back is only presented if there is at least one other supported MFA method to select.

On clients that don't support https://github.com/mautrix/go/pull/560, there is no change in behavior.

### DIFF
--- a/pkg/messagix/bloks/selenium.go
+++ b/pkg/messagix/bloks/selenium.go
@@ -339,7 +339,7 @@ type Browser struct {
 	AFADNotification string
 	AFADInterval     time.Duration
 	AFADCallback     func() error
-	AFADCanGoBack    bool
+	MFACanGoBack     bool
 
 	LoginData    string
 	DisplayedURL string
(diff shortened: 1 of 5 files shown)


pkg/messagix/bloks/selenium.go
```diff
@@ -1967,10 +1974,23 @@ func (b *Browser) DoLoginStep(ctx context.Context, userInput map[string]string)
 			if b.AFADCallback == nil {
 				return nil, loginerrors.AFADStopped
 			}
-			time.Sleep(b.AFADInterval)
+			timer := time.NewTimer(b.AFADInterval)
+			select {
+			case <-timer.C:
+			case <-ctx.Done():
+				if !timer.Stop() {
+					<-timer.C
```
tulir: This is no longer needed as of Go 1.23 https://go.dev/wiki/Go123Timer, just use a plain `time.After` in the select

tulir (approved):

PR #347, opened 2026-09-08
ig/login: add support for checkpoints in native login

highesttt:
The reason this diff is huge, is because this was not just one issue but a snowball effect of many issues back to back

- Fixed captcha not showing up
- Handles required cookie consent
- Handles expired cookies
- Added/Fixed SMS/email/Whatsapp checkpoint handling
- Support for the new ig verification screen
- Preserves device context when requesting verification codes
- Persists device identity
- handles recoverable errors

### Checklist


pkg/messagix/httpclient/http.go
```diff
@@ -338,6 +341,21 @@ func GetErrorRedirectURL(err error) string {
 	return ""
 }
 
+func unwrapURLErrors(err error) error {
+	for {
+		urlErr, ok := errors.AsType[*url.Error](err)
```
tulir: We don't use url.Error anywhere 🤔

pkg/messagix/httpclient/js_module_parser.go
```diff
@@ -105,6 +108,11 @@ func (m *ModuleParser) Load(ctx context.Context, page string) error {
 	if err != nil {
 		return err
 	}
+	if m.InspectPage != nil {
+		if err = m.InspectPage(htmlData); err != nil {
```
tulir, line 112: This looks hacky. Why isn't the parsing just done automatically like everything else?

pkg/metadb/14-instagram-usdid.sql
```diff
@@ -0,0 +1,6 @@
+-- v14 (compatible with v13+): Persist Instagram's signed installation identity
```
tulir: The compat version should usually not drift from the latest schema

```suggestion
-- v14 (compatible with v11+): Persist Instagram's signed installation identity
```

pkg/metadb/00-latest-schema.sql
```diff
@@ -98,6 +98,10 @@ CREATE TABLE meta_instagram_login_device (
 	advertising_id    TEXT NOT NULL,
 	android_device_id TEXT NOT NULL,
 	machine_id        TEXT NOT NULL,
+	usdid             TEXT NOT NULL,
+	usdid_key_id      TEXT NOT NULL,
+	usdid_private_key TEXT NOT NULL,
+	usdid_registered  BOOLEAN NOT NULL,
```
tulir: These should have the same `DEFAULT`s as the schema upgrade, otherwise it's not actually compatible with v13

pkg/messagix/httpclient/js_module_parser.go
tulir: Actually maybe it's fine as-is to avoid putting instagram web specific code with the rest of the module parser

tulir (approved):

PR #351, opened 2026-09-12
push,messagix: add native Messenger Android push registration

iFixRobots:
## Problem

Cookie-only Messenger sessions cannot register native encrypted push notifications.

## Fix

Prefer the Messenger Android login flow. Preserve native credentials, device identity and push keys, and register FCM with Graph and the encrypted connection. Keep the existing iOS cookie login behavior.

Handle native relay payloads through background sync, return failed or incomplete catch-up for retry, retain messages when chat sync fails, and reject invalid push data before indexing it.

## Notes

Existing cookie-only connections must log in again with Messenger Android. Beeper must promote `bridge:local-facebook:login:messenger-lite-android` for automatic selection; the existing iOS login flag otherwise takes precedence. Cached relay registrations must renew through the SDK's trusted re-registration event or normal expiry before selecting FCM. This uses existing client APIs.

Beeper relay support requires [push-receiver#4](https://github.com/beeper/push-receiver/pull/4), included in [hungryserv#588](https://github.com/beeper/hungryserv/pull/588).

Build and vet pass. A normal Pixel login registered its own push keys. A real encrypted provider push, forwarded with the existing relay serializer, produced a Beeper notification while the app was in the background. Replaying it after process death restarted the bridge without duplicate messages or notifications. These checks used a controlled relay forward; automatic server forwarding and iOS delivery still need release verification.

### DIFF
--- a/pkg/connector/login.go
+++ b/pkg/connector/login.go
@@ -2,6 +2,7 @@ package connector
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"maps"
@@ -9,7 +10,9 @@ import (
 	"slices"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/rs/zerolog"
+	waTypes "go.mau.fi/whatsmeow/types"
 	"maunium.net/go/mautrix/bridgev2"
 	"maunium.net/go/mautrix/bridgev2/database"
 	"maunium.net/go/mautrix/bridgev2/status"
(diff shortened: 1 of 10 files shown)


pkg/connector/login.go
```diff
@@ -221,22 +267,73 @@ func loginWithCookies(
...
+			authClient.MessengerLite.SetNativeSession(&updated)
+			newCookies, err := authClient.MessengerLite.ExchangeTransientToken(ctx, nativeSession.AccessToken)
+			if err != nil {
+				return nil, fmt.Errorf("failed to prepare native session for encrypted device: %w", err)
+			} else if newCookies.GetUserID() != id {
+				return nil, fmt.Errorf("exchanged native session account does not match login")
+			}
+			nativeSession = authClient.MessengerLite.GetNativeSession()
+		}
+	}
```
tulir: This part is a bit sus since NewLogin is meant to handle replacing existing logins. The entire block seems to mostly exist for some kind of partial session reuse? Is that even actually safe and is there some reason to do it instead of just replacing the entire session on relogin?

iFixRobots: Yeah... The intention was to keep push registration and encrypted messaging using the same device identity, but I realize now that that doesn't require preserving the old session during a fresh login.

tulir: Probably better to get rid of this whole block then and just let NewLogin use the default overwriting logic

pkg/connector/push.go
```diff
@@ -160,27 +184,38 @@ func (m *MetaClient) ensurePushMessageReceived(ctx context.Context, pd *pushcryp
...
 		log.Debug().
 			Str("message_id", msgID).
 			Str("chat_id", string(part.Room.ID)).
 			Stringer("event_id", part.MXID).
 			Msg("Confirmed push message was bridged after backfill")
 	} else {
-		log.Warn().
-			Str("message_id", msgID).
-			Msg("Push message still wasn't bridged after backfill")
+		return fmt.Errorf("push message %s wasn't bridged after backfill", msgID)
```
tulir: Hmm, returning an error for this might just cause loops where it tries handling the same push over and over even if it's never going to work

tulir (approved):
Looks good, other than that one error return which might be best to revert to a log

PR #359, opened 2026-09-17
ig/login: hand off unsupported AuthPlatform redirects

highesttt:
This pr is sort of a follow up to #358, known AuthPlatform pages can still redirect to ubsupported subroutes

### Checklist


Fizzadar (approved):

Fizzadar:
@highesttt I think https://github.com/mautrix/meta/pull/369 does the same thing but with less changes?

highesttt:
> @highesttt I think #369 does the same thing but with less changes?

Looked at it, they're different. #369 finishes login and redirects out of the flow while #359 uses the webview for unsupported AuthPlatform subroutes, kind of like #363 does, except that #359 also handles redirects that are encoutered later in the flow, which #363 does not. 

I think we should merge #359 and #369 and close #363

PR #361, opened 2026-09-18
ig: use one native login for messages and push notifications

iFixRobots:
Add a native Instagram login flow that uses one session for messages and push notifications. The flow is available for both cloud and on-device connections.

- Sign in once, including verification, and use that session for messaging and push registration.
- Register native push through Beeper's existing relay so Instagram can wake an on-device connection while Beeper is in the background.
- Save the session and push keys so reconnecting and renewing registration don't require another login.
- Reuse the existing message processing and sync handling. Keep existing web sessions working.
- Use req/uTLS for the native TLS and HTTP/2 profile, and the official FlatBuffers library for stream requests and responses.

Validation includes a fresh Android login, self-hosted login and restart, sending and receiving with native sessions, and background notifications on Android and iOS. Fresh iOS setup remains unverified because the SMS never arrived. A fresh install of Instagram's own Android app reproduced that nondelivery too.

Rollout requires enabling `instagram-native` for the cloud and local login flows.

Related: [SDK update](https://github.com/beeper/beeper/pull/4992) · [req receive-window fix](https://github.com/beeper/req/pull/2)


pkg/igconnector/client.go
```diff
@@ -121,13 +124,15 @@ func (ic *IGConnector) getProxy(reason string) (string, error) {
 func (ic *IGClient) ensureIGClient() {
 	if ic.LoginMeta.Cookies != nil && ic.Client == nil && ic.LoginMeta.Platform == types.Instagram {
 		ic.LoginMeta.Cookies.Platform = ic.LoginMeta.Platform
+		_, nativeMessaging := ic.Main.Bridge.Matrix.(bridgev2.MatrixConnectorWithNotifications)
```
tulir: There shouldn't be any checks like this, Android login should just be a new login flow

Fizzadar:
@highesttt can you test this out as well please? Isn’t this the login flow that wasn’t giving cookies that worked with the rest of the bridge (mobile only cookies or something)?

pkg/messagix/dgw/conn.go
```diff
@@ -0,0 +1,174 @@
...
+	reader   *bufio.Reader
+}
+
+func newHTTPConnection(ctx context.Context, opts *HTTPStreamOptions) (*connection, error) {
+	if opts.Client == nil {
+		return nil, fmt.Errorf("%w: HTTP client is required", ErrDial)
+	}
+	ctx, cancel := context.WithCancel(ctx)
+	input, output := io.Pipe()
+	request, err := http.NewRequestWithContext(ctx, http.MethodPost, opts.URL, input)
```
tulir, line 49: They're really using full duplex http/2 requests instead of just websockets (or even plain tcp)? Definitely more plausible than requiring http/3, but still seems slightly weird.

pkg/messagix/dgw/basestream.go
```diff
@@ -59,7 +58,11 @@ func newBaseStream(conn *websocket.Conn, id StreamID, log *zerolog.Logger) baseS
 }
 
 func receiveEstablish(f *EstablishStreamFrame, log *zerolog.Logger) error {
-	if res := gjson.GetBytes(f.RawParameters, "code"); !res.Exists() {
+	res := gjson.ParseBytes(f.RawParameters)
+	if res.IsObject() {
+		res = res.Get("code")
+	}
+	if !res.Exists() || res.Type != gjson.Number {
```
tulir, line 65: What's the point of this change? Doesn't seem to do anything useful

pkg/instameow/socket.go
```diff
@@ -172,6 +172,12 @@ func (c *Client) getSocketOptions() dgw.SocketOptions {
 			return err
 		},
 	}
+	if c.mobileSession != nil {
+		options.HTTPStream = &dgw.HTTPStreamOptions{
+			Client: c.http.HTTP, URL: "https://test-gateway.instagram.com/lightspeed", GetHeaders: c.nativeSocketHeaders,
```
tulir: test-gateway?

pkg/messagix/httpclient/instagram_tls.go
```diff
@@ -0,0 +1,71 @@
...
+func setInstagramTLSFingerprint(client *req.Client) {
+	fallback := client.GetTransport().TLSHandshakeContext
+	client.SetTLSHandshake(func(ctx context.Context, addr string, plainConn net.Conn) (net.Conn, *tls.ConnectionState, error) {
+		host := addr
+		if name, _, err := net.SplitHostPort(addr); err == nil {
+			host = name
+		}
+		if host != "i.instagram.com" {
+			return fallback(ctx, addr, plainConn)
+		}
```
tulir: Why is this limited to i.instagram.com? 🤔 surely the mobile app would use roughly the same fingerprint for all requests?

(calling this entire function should probably be gated to the mobile credentials instead)

pkg/instameow/native_socket.go
```diff
@@ -0,0 +1,95 @@
...
+	userID := c.GetOwnFBID()
+	if userID == 0 {
+		return nil, fmt.Errorf("native messaging user ID is missing")
+	}
+	return marshalNativeStreamRequest(uint16(retryCount%65535+1), uint64(userID), payload), nil
+}
+
+func marshalNativeStreamRequest(requestID uint16, userID uint64, payload []byte) []byte {
+	b := make([]byte, 56, 61+len(payload))
+	binary.LittleEndian.PutUint32(b, 24)
```
tulir: Extracting the real flatbuffer schema would be preferable to manual byte handling, though I guess it might be fine if this is the only place it happens

pkg/instameow/login_mobile.go
```diff
@@ -338,7 +345,7 @@ func (c *Client) mobileLoginHeaders(state *mobileLoginState) http.Header {
 }
 
 func (c *Client) makeMobileLoginRequest(ctx context.Context, state *mobileLoginState, requestURL string, headers http.Header, body []byte) ([]byte, error) {
-	response, responseBody, err := c.http.MakeRequest(ctx, requestURL, http.MethodPost, headers, body, types.FORM)
+	response, responseBody, err := c.http.MakeRequestOnce(ctx, requestURL, http.MethodPost, headers, body, types.FORM)
```
tulir: Why is this Once? Generally we should let auto-retries do their thing unless there's a specific reason why a request can't be retried ever

pkg/instameow/login_web.go
```diff
@@ -417,6 +419,11 @@ func (c *Client) captureInstagramWebTwoFactor(
 		Bool("csrf_rotated", responseCSRFToken != "" && responseCSRFToken != preResponseCSRFToken).
 		Bool("csrf_retained", retainedCSRFToken).
 		Msg("Captured Instagram web two-factor challenge")
+	if info.EncryptedContext != "" && (method == "SMS" || method == "WHATSAPP") {
+		if err := c.sendInstagramWebTwoFactorCode(ctx, c.webTwoFactor); err != nil {
+			return nil, err
```
tulir, line 424: Was this something actually missing from the web login code, or was it just for the mixed native+web login thing that was removed? (also are there any other leftovers of that mixed login flow?)

pkg/instameow/native_socket.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	offset := table.Offset(6)
+	if offset == 0 {
+		return nil, nil
+	}
+	field := uint64(table.Pos) + uint64(offset)
+	relative := uint64(flatbuffers.GetUOffsetT(b[field:]))
+	if relative == 0 || field+relative+4 > uint64(len(b)) {
+		return nil, invalid
+	}
```
tulir: What's all this for 🤔

pkg/messagix/bloks/bloks.go
```diff
@@ -14,8 +14,7 @@ const BloksVersionIOS = "e74cbfe42b6152a793f3ee3d81eeba4c48b5c4477b54d05ab8d8bce
 // Messenger Android 569.0.0.44.91 of 2026-07-22
 const BloksVersionAndroid = "194a25b5ca64b7e2cc9a1b57a306ae5b1536d14d53e54f273f19a639c21cb197"
 
-// Instagram Android 440.0.0.19.86 of 2026-07-24.
-const BloksVersionInstagramAndroid = "3ae1e6445cdefeda05bde09f57499d84157ad74699044bf7fe9ca9183150f17a"
+const BloksVersionInstagramAndroid = "935a519904e9017324cdedb64a283a3c2c1a3d5b0bbc698b451f5aef72cc11df"
```
tulir, line 18: We should keep a source version comment

pkg/messagix/dgw/conn.go
tulir: The apk does seem like it has no websocket even as a fallback and always uses a full duplex http/2 (or http/3) post request

iFixRobots:
1. **Was the SMS request missing from web login? Any mixed-login leftovers?**

It was missing from web login too. I found it during the mixed-flow work, but Instagram’s website makes a separate request to send the first SMS. We were showing the code field without making that request, so that fix stays.

There were leftovers I should have removed. Those are gone now. 

2. **What were the manual FlatBuffers checks for?**

I kept them after switching to the generated decoder and missed that the surrounding DGW handler already catches decoding failures and reconnects. I removed the whole wrapper. It now uses the generated decoder directly. 

3. **Keep the source-version comment.**

Agreed and done.

tulir:
I renamed the new login flow to `android` since `instagram-native` will be even more confusing if we ever add ios (and there's no need for the `instagram-` prefix anymore with IG and FB fully split up)

Everything mostly looks good now, but @highesttt should also review the login parts

highesttt (approved):
Looks reasonable, tested on multiple accounts from different countries, works well

